PR #4, opened 2026-06-22
feat: improvements for fabric validate CLI command

dagardner-nv:
* Expand the validate sub command to ensure that the profile YAML files are valid YAML, fix the help strings for the validate command.
  * This just runs `discover_profiles` and catches anything that would be caught by `fabric run` when a profile appears in the profiles dir, but isn't currently selected.
* Catch duplicate profile names rather than silently allowing the second profile to win
* Log a warning for any missing profile directories
* Ensure the `.tmp` directory is created on initial clone

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ tests/fixtures/*/artifacts/
 # UV Lock
 # TBD if we want to track this in git
 uv.lock
+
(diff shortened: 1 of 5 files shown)


AjayThorve (approved):
Very helpful!

PR #14, opened 2026-06-23
Rename Hermes adapter setting max_turns -> max_iterations

AjayThorve:
Follow-up to #12. Renames the Hermes adapter setting `max_turns` → `max_iterations`.

## Why
`max_turns` maps to Hermes' `AIAgent(max_iterations=…)` — the cap on agent-loop iterations within a single invocation. The name reads as a conversational-turn count, which is misleading (a session "turn" is one `invoke()`). `max_iterations` matches Hermes' own parameter.

## What
- Rename the setting in the `hermes-sdk`, `hermes-session`, `hermes-cli`, and `hermes-relay` profiles.
- Both adapters read `settings.get("max_iterations")`.
- Hermes' config-file key (`agent.max_turns`) and the SDK parameter (`max_iterations`) are unchanged; runtime behavior is identical.

## Tests
- All four profiles resolve with `max_iterations: 1`; adapter syntax + profile YAML verified. The setting only reaches Hermes on gated real-Hermes runs.

## Summary by CodeRabbit

* **Configuration Updates**
  * Standardized configuration parameter naming across Hermes adapters for consistency and clarity. The iteration limit setting is now referred to as `max_iterations` instead of `max_turns` in all Hermes implementations. Users with existing Hermes configurations should update their settings to use the new parameter name.

### DIFF
--- a/adapters/hermes-sdk/src/nemo_fabric_adapters/hermes_sdk/adapter.py
+++ b/adapters/hermes-sdk/src/nemo_fabric_adapters/hermes_sdk/adapter.py
@@ -254,7 +254,7 @@ def build_hermes_config(payload: dict[str, Any], *, relay_enabled: bool = False)
         ),
         "agent": without_none(
             {
-                "max_turns": settings.get("max_turns"),
+                "max_turns": settings.get("max_iterations"),
                 "disabled_toolsets": settings.get("disabled_toolsets"),
             }
         ),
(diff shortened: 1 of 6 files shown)


adapters/hermes-sdk/src/nemo_fabric_adapters/hermes_sdk/adapter.py
```diff
@@ -254,7 +254,7 @@ def build_hermes_config(payload: dict[str, Any], *, relay_enabled: bool = False)
         ),
         "agent": without_none(
             {
-                "max_turns": settings.get("max_turns"),
+                "max_turns": settings.get("max_iterations", settings.get("max_turns")),
```
dagardner-nv: Shouldn't this be?
```suggestion
                "max_iterations": settings.get("max_iterations", settings.get("max_turns")),
```

dagardner-nv (approved):

adapters/hermes-sdk/src/nemo_fabric_adapters/hermes_sdk/adapter.py
AjayThorve: unfortunately, hermes uses `max_turns` in it's own config.yaml, but `max_iterations` via hermes constructor.

So we anyways need to normalize fabric's input to whatever hermes expects.

PR #22, opened 2026-06-26
Expand Relay plugin configuration

dagardner-nv:
* The Hermes Relay plugin supports configuring both ATOF and ATIF Relay plugins from environment variables.
* While other Relay plugins, notably OTel and OpenInference, require a Relay TOML configuration file or a `PluginConfig`
* Add a new `write_relay_plugins_toml` helper method for hermes_cli
* Add new `_relay_api_plugin_config` helper method for hermes_sdk
* Add a new `adapters/common/src/nemo_fabric_adapters/common/utils.py` module for methods like `write_relay_plugins_toml` which are not Hermes specific and could be of use to other adapters.
* Split `tests/test_adapaters_hermes_common.py` tests into those for the common/hermes and common/utils tests
* Always use the `hermes chat --query <prompt>` command in the `hermes_cli` adapter rather that the `-z` flag, the reason is that with `-z` the session is not finalized on completion the way it is with `hermes chat`, this conforms with behavior of the `hermes_sdk` adapter.
* Remove require_yaml parameter, since JSON is a subset of YAML, writing JSON to a YAML file is always valid we should not raise exceptions on a missing PyYaml library.
* Move the `pyyaml` dependency to be a dependency of the `hermes` extra (currently only need to write Hermes config files)
* Add a dependency for `tomli-w` to the `relay` extra

## Summary

- **New Features**
  - Added Hermes Relay/OpenInference and Relay/OpenTelemetry example profiles for oneshot tracing and artifact generation.
  - When relay is enabled, Hermes config generation is standardized and may emit a Relay plugins TOML for plugin setup.

- **Bug Fixes**
  - Updated Hermes CLI argument handling to consistently use `chat --quiet --query <prompt>` (including session continuation).

- **Refactor**
  - Unified shared Hermes/relay config and payload normalization across the Hermes CLI and SDK adapters.

- **Tests**
  - Updated and expanded unit, smoke, and end-to-end tests to match the new shared relay/config behavior and CLI argument expectations.

### DIFF
--- a/adapters/common/src/nemo_fabric_adapters/common/hermes.py
+++ b/adapters/common/src/nemo_fabric_adapters/common/hermes.py
@@ -7,9 +7,10 @@ from __future__ import annotations
 
 import json
 import os
-import sys
 from pathlib import Path
 from typing import Any
+
+from nemo_fabric_adapters.common.utils import write_relay_plugins_toml
 
 
 def effective_config(payload: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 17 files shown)


AjayThorve:
@dagardner-nv I think we should split the Relay configuration path by adapter boundary.

For `hermes-cli`, generating a per-invocation Relay plugin TOML is the correct path. Fabric is launching Hermes as a separate process, so the Relay config has to cross a process boundary through process-visible state: env, CLI args, or a file. Relay’s own plugin config discovery is file-based (`plugins.toml` layering), so TOML is a reasonable bridge there.

For `hermes-sdk`, though, we are already in the same Python process before Hermes plugin discovery / agent construction. In that path we should avoid writing `relay-plugins.toml` and avoid mutating process-global `HERMES_NEMO_RELAY_*` variables. Relay exposes an in-process API for this:

```python
from nemo_relay import plugin
from nemo_relay.observability import (
    AtofConfig,
    AtifConfig,
    ComponentSpec,
    ObservabilityConfig,
    OtlpConfig,
)

plugin_config = plugin.PluginConfig(
    components=[
        ComponentSpec(
            ObservabilityConfig(
                atof=AtofConfig(...),
                atif=AtifConfig(...),
                openinference=OtlpConfig(...),
            )
        )
    ]
)

async with plugin.plugin(plugin_config):
    # run Hermes SDK conversation
    ...

AjayThorve (approved):
This split looks right to me now: CLI keeps serialized TOML/env because it launches Hermes out-of-process, and SDK uses the in-process Relay plugin API with scoped cleanup.

Non-blocking follow-up: it would still be useful to add an SDK regression test that exercises a Relay-enabled run followed by a Relay-disabled run in the same Python process, but I don’t think that needs to block this PR.

PR #30, opened 2026-07-01
feat: Add Relay integration to codex-cli adapter

dagardner-nv:
* Spin up `nemo-relay` in gateway mode independently of Codex
* Write out a temporary Codex config profile and remove it at the end of the run (unfortunately this needs to be in `~/.codex`)
* Consolidate profiles so that they stack on each other ex:
  ```sh
  fabric run examples/code-review-agent \
    --profile hermes_sdk \
    --profile relay_otel \
    --input "who are you?"
  ```
* Move `load_payload` and `runtime_session_id` to `common/utils.py`
* `write_relay_plugins_toml` now writes both a config.toml and a plugins.toml, the method was renamed to `write_relay_configs`

## Summary by CodeRabbit

* **New Features**
  * Added optional Codex routing via NeMo Relay gateway, including automatic gateway startup/shutdown and generated Codex/Relay profile handling.
  * Added new telemetry profiles for Relay (OpenTelemetry and OpenInference).
* **Bug Fixes**
  * Improved payload input loading (invocation file or stdin fallback).
  * Made session identifier derivation consistent across Hermes components.
  * Enhanced Relay TOML config generation and wiring.
* **Tests**
  * Expanded smoke, unit, and E2E coverage to support multiple profiles per run and validate gateway/profile lifecycle.


adapters/common/src/nemo_fabric_adapters/common/utils.py
```diff
@@ -312,21 +334,36 @@ def collect_relay_artifacts(plugin_config: dict[str, Any]) -> list[dict[str, str
     return artifacts
```
AjayThorve, line 337: `collect_relay_artifacts()` returns every matching file already present in the shared Relay output directory, rather than only artifacts produced by the current invocation.

Repro:

start with clean artifacts and run `codex_cli + relay` twice. The second `RunResult` contains both the first run’s ATIF and the second run’s ATIF; both are also promoted into the second `ArtifactManifest` as `relay_atif` and `relay_atif_2`. This breaks per-invocation artifact isolation and can feed stale trajectories into evaluation or upload consumers.

ideally Relay output should be scoped by runtime/invocation

AjayThorve (approved):
The intended Codex + Relay flow works end to end, including gateway lifecycle and ATOF/ATIF generation. Requesting changes only for the cross-run artifact contamination above.

PR #36, opened 2026-07-07
feat(adapters): resolve built-in adapters from an installed distribution

SandyChapman:
## Problem

Built-in adapters resolve only from two filesystem sources: the compile-time repository tree (`env!("CARGO_MANIFEST_DIR")/../../adapters` — an absolute path baked into the binary) and `<agent_root>/adapters`. Neither is relocatable, and the wheel shipped no descriptors, so an **installed** `nemo-fabric` (or a moved binary) cannot resolve `nvidia.fabric.codex.cli` without a repo checkout — callers must pass `base_dir=<checkout>`. This blocks using the SDK from an arbitrary directory (notebooks, CI) and any future distributed wheel.

## Change

Make built-in adapters travel with the installed package and resolve with **no `base_dir`, no env var, and no Python import side effect**:

- **Move the maintained adapters under the Python package** (`adapters/` → `python/src/nemo_fabric/adapters/`). They are already Python (`nemo_fabric_adapters` packages + self-locating `adapter.py`), so this is their natural home, and maturin bundles files under `python-source` into the wheel automatically. The compile-time `repository_adapter_dir()` is repointed at the new location for in-repo dev/CLI builds.
- **Add `AdapterDescriptorSource::Installed`.** `AdapterRegistry::from_config` registers the bundle discovered at runtime beside the loaded native extension (`process_path::get_dylib_path()` → `<dylib dir>/adapters`), between `Repository` and `Local`. Precedence (last-writer): **Local > Installed > Repository**.

Net effect: `pip install nemo-fabric[codex]` + the `codex` binary resolves and runs the codex adapter with nothing set.

## Why the move (vs. keeping `adapters/` top-level)

maturin only bundles files that are **tracked under `python-source`**: `[tool.maturin] include` lands files at their top-level path (`site-packages/adapters/`, not under the package), and a `build.rs` copy into a git-ignored dir is excluded because maturin honors `.gitignore`. Moving keeps a single source of truth with no build step or tracked duplicate. (Alternative if reviewers prefer `adapters/` stay top-level: commit a vendored copy under `python/src/nemo_fabric/adapters/` synced by a build step — trades the move for tracked duplication.)

The `process_path` dep wraps `dladdr`/`GetModuleFileNameW`; if you'd rather not add a dep, the same self-location can be hand-rolled with `libc`/`windows-sys` (already in the tree).

## Testing

- `cargo test -p fabric-core` — 46 passed (incl. new Installed-over-Repository precedence test); RunPlan schema snapshot regenerated for the new enum value.
- Built the wheel and verified it ships `nemo_fabric/adapters/**`.
- Real `codex` run from the installed wheel with `base_dir=None` and nothing set → `status: succeeded`.

## Notes

Draft — `cargo fmt`/`clippy` not run locally (components unavailable in my environment); CI will cover. Opening from a fork (read-only on upstream).

## Summary by CodeRabbit

* **New Features**
  * Expanded adapter discovery to include adapters bundled with the installed Fabric distribution and discovered at runtime.
  * Added `"installed"` as a supported adapter descriptor source in the run-plan schema.

* **Bug Fixes**
  * Improved precedence so, when an adapter exists in multiple locations, the installed/bundled version overrides repository-provided entries.

* **Documentation**
  * Clarified that maintained harness adapters are packaged inside the Python distribution, and harness-specific extensions should be configured via `harness.settings`.


AnuradhaKaruppiah:
This PR was discussed offline with Sandy and we will be providing wheels per-adapter. Sandy, Is it OK to close this PR?

PR #45, opened 2026-07-09
Add support for an `ADAPTER_PYTHON` environment variable as a shortcut for setting `python_env` in the config

dagardner-nv:
#### Overview

* Expands on the `HERMES_PYTHON` idea used in the Hermes Agent example
* `ADAPTER_PYTHON` functions as the default value for the `python_env` setting, when unset follows existing fallback rules.
* CLI adapters now determine if they are in a virtualenv and set `PATH` and other environment variables to ensure sub-processes are run within the virtualenv.

Unrelated changes:
* Add a `--show-output` flag to `examples/code_review_agent/__main__.py`, to make reading the output easier
* Ensure that the packages are all installed as editable

`crates/fabric-core/src/runtime.rs`

- Closes FABRIC-56

## Summary by CodeRabbit

* **New Features**
  * Added a `--show-output` option to display the adapter response after the normalized result.
  * Added `ADAPTER_PYTHON` support for selecting the adapter Python interpreter.

* **Bug Fixes**
  * Added preflight validation for the adapter interpreter path (must exist and be a file).
  * Refined interpreter precedence and fallback behavior across harness settings and `ADAPTER_PYTHON`.
  * Improved adapter subprocess environment handling to better honor active virtual environments.

* **Documentation**
  * Updated quick start and example instructions to use `ADAPTER_PYTHON` and clarified precedence.

* **Chores / Tests**
  * Updated editable install configuration and expanded coverage for adapter selection and environment behavior.


adapters/codex-cli/src/nemo_fabric_adapters/codex_cli/adapter.py
```diff
@@ -488,15 +458,20 @@ def build_env(
     *,
     relay_gateway_url: str | None = None,
 ) -> dict[str, str]:
-    env = {name: os.environ[name] for name in INHERITED_ENV_NAMES if name in os.environ}
+    env = common_utils.virtualenv_subprocess_env()
```
ericevans-nv, line 461: This changes the Codex adapter from an explicit env allowlist to forwarding the full parent environment via `virtualenv_subprocess_env()`. That can expose unrelated secrets/tokens to `codex exec`. Is this intentional?

examples/code_review_agent/__main__.py
```diff
@@ -49,6 +54,11 @@ async def main() -> None:
     else:
         output = await fabric.run(config, base_dir=BASE_DIR, input=args.input)
     print(json.dumps(output.to_mapping(), indent=2))
+    if args.show_output and not args.plan:
+        if isinstance(output.output, dict) and "response" in output.output:
+            print(f"\n{output.output['response']}")
+        else:
+            print(f"\n{output.error.message}")
```
ericevans-nv, line 61: `--show-output` can raise on valid successful results that do not expose `output.response`. `RunResult.error` is `None` on success, so this fallback can hit `output.error.message` and fail with `AttributeError` even though the run itself succeeded. This should handle the no-error case instead of assuming a missing `response` means there is an error object.

adapters/codex-cli/src/nemo_fabric_adapters/codex_cli/adapter.py
dagardner-nv: Yes, it's intentional, this was behavior specific to only this adapter.

examples/code_review_agent/__main__.py
dagardner-nv: Currently all three of the adapters define an `output.response` field, but you're right, it isn't defined in the core lib as one of the output fields. 

I would argue it should be.

ericevans-nv, line 61:
```suggestion
    if args.show_output and not args.plan:
        if isinstance(output.output, dict) and "response" in output.output:
            print(f"\n{output.output['response']}")
        elif output.error is not None:
            print(f"\n{output.error.message}")
        else:
            print("\n(run succeeded but output has no 'response' field)")
```

ericevans-nv (approved):
Approving since the core ADAPTER_PYTHON behavior looks good and the Codex env inheritance is intentional. I left one suggested guard for --show-output so it doesn’t raise on a successful result that lacks output.response; that can be applied here or handled as a follow-up if you want to formalize response as part of the output contract separately.

PR #54, opened 2026-07-13
feat(relay): add typed Relay config

AnuradhaKaruppiah:
#### Overview

Adds first-class Relay config as a top-level `relay` block and changes telemetry selection to a provider map:

```yaml
telemetry:
  providers:
    relay: {}
    native:
      config: {}
```

Configured providers are enabled providers. Relay-specific project, output, observability, components, and policy live under the top-level `relay` block.

#### Relay sample

```yaml
telemetry:
  providers:
    relay: {}

relay:
  project: code-review-smoke
  output_dir: ./artifacts/relay
  observability:
    atof:
      enabled: true
    atif:
      enabled: true
```

Start with `crates/fabric-core/src/config.rs` for the source-of-truth config and telemetry plan, then `python/src/nemo_fabric/models.py` / `python/src/nemo_fabric/types.py` for SDK authoring, and `adapters/common/src/nemo_fabric_adapters/common/utils.py` for adapter-facing helpers.

#### Adapter guidance

- Treat `telemetry_plan` as the adapter contract; do not infer telemetry intent from `effective_config.config.telemetry`.
- Enable Relay only when `telemetry_plan.relay_enabled` is `true`.
- Read native telemetry settings from `telemetry_plan.native_config`.
- Treat `FABRIC_RELAY_CONFIG_PATH` and related Relay env vars as runtime state, not as enablement signals.

#### Follow-up

- Narrow `nemo_fabric.__init__` exports in a separate PR: keep SDK entrypoints/errors at the package root, and import config/result types from `nemo_fabric.models` and `nemo_fabric.types`.

- CLOSES FABRIC-76

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project license.

## Summary by CodeRabbit

* **New Features**
  * Added first-class NeMo Relay configuration with typed observability, components, exports, storage, and validation settings.
  * Telemetry now supports multiple providers, including native telemetry pass-through configuration.
  * Added provider management APIs for enabling and removing telemetry providers.

* **Bug Fixes**
  * Improved telemetry planning and metadata when no providers are configured.
  * Relay enablement and configuration are now resolved consistently from their dedicated configuration sections.

* **Documentation**
  * Updated SDK, schema, API reference, examples, and configuration guides for the new telemetry and Relay formats.


Salonijain27 (approved):
Approved from a dependency point of view

crates/fabric-core/src/config.rs
```diff
@@ -1271,18 +1678,71 @@ fn resolve_telemetry_plan(
...
+            .then(|| relay.and_then(|relay| relay.project.clone()))
+            .flatten(),
+        relay_output_dir: relay_enabled
+            .then(|| relay.and_then(|relay| relay.output_dir.clone()))
+            .flatten(),
+        relay_config: relay_enabled
+            .then(|| resolve_relay_plugin_config(relay))
+            .flatten(),
+        native_config: native_provider.and_then(|provider| provider.config.clone()),
         adapter_outputs: adapter_descriptor
```
AjayThorve: this copies the adapter's declared outputs into the plan but never validates the selected providers against the adapter's capabilities.

I can enable Relay for the current Claude adapter and `fabric plan` succeeds even though its descriptor declares no telemetry support and the adapter implements no Relay path. The same mismatch exists in the opposite direction for Codex, which has Relay code but does not declare it.

Can we have adapter declare what they support using a simple contract which would look like:

```
{
  "telemetry": {
    "providers": {
      "relay": {
        "outputs": ["atif", "otel", "openinference"],
        "integration_modes": ["hooks", "gateway"]
      },
      "native": {
        "outputs": ["otel"]
      }
    }
  }
}
```

the `resolve_telemetry_plan` can then take this and run a check like `!adapter.telemetry.supports_provider(requested_provider)` or something similar to validate. This would atleast reject unsupported adapter/provider combinations during plan resolution. 

Thoughts?

python/src/nemo_fabric/models.py
```diff
@@ -355,16 +477,30 @@ def enable_relay(
...
+        relay_observability = (
+            observability
+            if observability is None or isinstance(observability, RelayObservabilityConfig)
+            else dict(observability)
+        )
+        relay_components = [
+            item if isinstance(item, RelayComponentConfig) else dict(item) for item in (components or [])
+        ]
+        relay_policy = policy if policy is None or isinstance(policy, RelayConfigPolicy) else dict(policy)
+        self.relay = RelayConfig(
```
AjayThorve: more a nitpick really, but doesn't this unconditionally replace the existing `RelayConfig`, so calling `enable_relay()` more than once erases previously configured values when the later call omits them.

For example:

```python
config.enable_relay(
    project="fabric",
    output_dir="./relay",
)
config.enable_relay()
```

After the second call, both `project` and `output_dir` are reset to None. I reproduced this against the current PR head.

Should we preserve the existing `self.relay` value and update only fields explicitly supplied by the caller? At minimum, calling enable_relay() with no arguments on an already-enabled configuration should be idempotent.


this is making an assumption on the usage of the `enable_relay` helper though, feel free to push back

crates/fabric-core/src/config.rs
```diff
@@ -604,29 +609,399 @@ pub enum McpExposure {
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
+pub struct RelayAtofEndpointConfig {
+    /// Endpoint URL.
+    pub url: String,
+    /// Endpoint transport.
+    #[serde(default)]
+    pub transport: RelayAtofEndpointTransport,
+    /// Endpoint headers.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub headers: BTreeMap<String, String>,
```
AjayThorve, line 757: Not fully aware of relay's lifecycle, but could the headers contain authorization related secrets as well? should we declare any secret handling for this case?

python/src/nemo_fabric/models.py
AnuradhaKaruppiah: aah, yes idempotency is important. but in that case the expectation is that API is executed in same fashion, right?

crates/fabric-core/src/config.rs
AnuradhaKaruppiah: I don't have enough info to answer this question yet. shall we pick this up in your PR?

AjayThorve: yeah, not a blocker by any means

crates/fabric-core/src/config.rs
AnuradhaKaruppiah: yes, good check to add. I will take care of it.

python/src/nemo_fabric/models.py
AjayThorve: Yes, fair distinction. Repeating the same call with the same arguments already produces the same state; my no-argument example is more about preservation/composability than strict idempotency.

AjayThorve: the reason i noticed this is `enable_native()` preserves existing native config when called without a new `config`.

`enable_relay()` starts from the existing Relay mapping and updates only non-`None` fields.

AnuradhaKaruppiah: on second thought I think people expect "additive" semantics i.e. unless explicitly cleared with a None/empty value the previous value is preserved. we can make that the rule of thumb. I will address this review comment

crates/fabric-core/src/config.rs
AnuradhaKaruppiah: Deferred

python/src/nemo_fabric/models.py
AnuradhaKaruppiah: addressed

PR #64, opened 2026-07-14
fix(hermes): default `max_iterations` to Hermes' default

yczhang-nv:
## Summary

The Hermes adapter constructed `AIAgent` with `max_iterations` defaulting to **1** when `harness.settings.max_iterations` was unset (`adapters/hermes/.../adapter.py`), giving the agent a single model turn. Multi-step tasks ran out of budget before producing an answer, yet the trial still reported `succeeded` — a **silently-wrong result**, not an error.

This changes the default to Hermes' own `AIAgent` default (**90**) via a named constant `DEFAULT_MAX_ITERATIONS`, so the adapter no longer overrides a sane default with a starving one.

## Changes

- `adapters/hermes/src/nemo_fabric_adapters/hermes/adapter.py`: add `DEFAULT_MAX_ITERATIONS = 90` (mirrors Hermes' `agent/agent_init.py` default) and use it at the `AIAgent(...)` call site instead of the literal `1`.
- `tests/adapters/test_hermes_adapter.py`:
  - Update the runtime-mapping test to expect `max_iterations=90` when unset (regression guard for the "unconfigured → not 1" path).
  - Add `test_default_max_iterations_matches_hermes_library_default` — cross-checks the constant against Hermes' own `AIAgent.__init__` default so drift is caught.
  - Add `test_build_hermes_config_omits_max_turns_when_max_iterations_unset` — documents that the config layer leaves `agent.max_turns` to Hermes when unset.

## Testing

`uv run --no-sync pytest tests/adapters/test_hermes_adapter.py` → 18 passed.

## Notes / follow-ups (tracked in FABRIC-85)

- If a low iteration cap is ever intentional as a cost guard, it should be documented and emit a one-line log when the loop terminates on the cap, so "hit the budget" is distinguishable from "the model chose to stop."
- Consider surfacing "hit iteration limit" as a degraded (non-`succeeded`) outcome so a truncated run doesn't read as a clean pass downstream.

Fixes FABRIC-85.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Hermes adapter to use a higher default agent iteration budget (90) when no custom limit is provided.
  * Treated an explicitly unset iteration setting as “not provided” to avoid conversion issues, while keeping turn-limit fields omitted in generated configuration.
* **Tests**
  * Added/updated Hermes regression tests to validate the new default iteration budget and configuration omission behavior, plus runtime construction expectations.

### DIFF
--- a/adapters/hermes/src/nemo_fabric_adapters/hermes/adapter.py
+++ b/adapters/hermes/src/nemo_fabric_adapters/hermes/adapter.py
@@ -26,7 +26,7 @@ import nemo_fabric_adapters.common.utils as common_utils
 # Mirrors Hermes' own AIAgent default (agent/agent_init.py); a lower value such
 # as 1 silently starves multi-step tasks (they run out of budget before
 # answering while the trial still reports success). See FABRIC-85.
-DEFAULT_MAX_ITERATIONS = 90
+DEFAULT_MAX_ITERATIONS: int = 90
 
 
 def validate_hermes_telemetry_provider(payload: dict[str, Any]) -> None:
@@ -284,6 +284,10 @@ def _invoke_hermes(
         session_id = common_utils.runtime_id(payload)
         session_db = SessionDB()
         conversation_history = load_runtime_history(session_db, session_id)
+        # Treat an explicit null max_iterations like an unset one (avoid int(None)).
+        max_iterations = settings.get("max_iterations")
+        if max_iterations is None:
+            max_iterations = DEFAULT_MAX_ITERATIONS
         agent = None
         agent = AIAgent(
             **filter_supported_kwargs(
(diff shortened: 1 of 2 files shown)


AnuradhaKaruppiah (approved):

yczhang-nv:
Addressed the CodeRabbit review in fd5efee (all three findings):

- **Type annotation** — `DEFAULT_MAX_ITERATIONS: int = 90`.
- **Explicit-null handling** — a `max_iterations: null` in `harness.settings` is now treated like an unset value and resolves to the default, instead of raising `int(None)` `TypeError`. (`settings.get(key, default)` only returns the default when the key is *absent*, so an explicit `null` previously slipped through to `int(None)`.)
- **Tests** — added `test_build_hermes_config_omits_max_turns_when_max_iterations_null` for the config layer, and fed `max_iterations: null` through the existing runtime-mapping test so the `AIAgent` path's null handling is regression-covered (it asserts the resolved value is `90`, which fails under the old `int(None)`).

**Validation**: `pytest tests/adapters tests/python` → 253 passed; copyright pre-commit hook clean.

PR #65, opened 2026-07-14
fix(runtime): prefer `venv python` before bare `python3`

yczhang-nv:
## Summary

A `python`-kind adapter was launched with a final fallback to **bare `python3` off PATH**. When Fabric ran from a project venv whose PATH `python3` was a *different* interpreter (e.g. Homebrew 3.13 vs. the venv's 3.11), the preinstalled adapter — installed in the caller's venv — was launched under the wrong interpreter and died mid-run with an opaque `ModuleNotFoundError: No module named 'nemo_fabric_adapters'`. Preflight never validated the fallback, so the mismatch only surfaced as a subprocess crash.

## Changes

`crates/fabric-core/src/runtime.rs`:
- **`resolve_python_command_with_env`** now prefers, before bare `python3`: `$VIRTUAL_ENV/bin/python` → an interpreter next to the running Fabric executable → `python3`. This only affects the *no-explicit-interpreter* fallback path; `harness.settings.python`, `settings.python_env`, and `ADAPTER_PYTHON` still take precedence exactly as before.
- Each resolved interpreter is tagged with its provenance via a new `PythonSource` enum, used for diagnostics.
- **`validate_python_command`** now validates **every** branch at preflight (previously only `ADAPTER_PYTHON` was checked): concrete paths must be an existing file; bare command names must resolve on `PATH`. Failures raise a new `FabricError::PythonInterpreterUnavailable` that names where the interpreter came from and how to override it, instead of an opaque mid-run crash.

`crates/fabric-core/src/error.rs`:
- Replaced the narrow `InvalidAdapterPython` variant with the general `PythonInterpreterUnavailable { path, origin, reason }`.

## Why this is safe

- The venv/host preference activates **only** when no interpreter is explicitly configured, so it never overrides user intent.
- The env-inherited child already carries `VIRTUAL_ENV`; this change makes the launcher *select* the matching interpreter rather than ignoring it (which is what the Python common layer already assumes).
- `python_adapter_runs_hermes_shim` (which actually spawns the interpreter) still passes: the fixture is made importable via `PYTHONPATH`, so any valid Python 3 works.

## Testing

`cargo test -p fabric-core` → **55 passed**. New/updated unit tests cover: virtualenv preference, host-interpreter preference, provenance tagging, and preflight rejection of a missing resolved interpreter. `cargo fmt --all --check` clean.

## Notes / follow-ups (tracked in FABRIC-86)

- Documenting `harness.settings.python` / `ADAPTER_PYTHON` as the escape hatch is deferred to the settings-discoverability work (FABRIC-84).

Fixes FABRIC-86.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Python adapter interpreter discovery across platforms with a clearer fallback order: active virtual environment first, then an adjacent installation, then system defaults.
  * Added stricter preflight validation to reject missing or invalid interpreters before startup.
  * Treats unset/empty interpreter configuration values as not configured, preventing accidental “empty path” failures.
  * Provides more actionable diagnostics when the resolved interpreter can’t be used.
* **Documentation**
  * Updated public Rust API docs and user-facing error messaging, including guidance to set a valid interpreter via settings or `ADAPTER_PYTHON`.
* **Tests**
  * Expanded tests for override, fallback behavior, and preflight rejection scenarios.


yczhang-nv:
Addressed the CodeRabbit review in 9ccb674:

**Fixed**
- **Empty env values treated as unset** — a set-but-empty `python_env` target or `ADAPTER_PYTHON` now falls through to the shared fallback chain instead of yielding an empty interpreter path.
- **Virtualenv resolves even when missing** — inside an active `VIRTUAL_ENV`, the resolver now always targets `<venv>/bin/python`, so a missing interpreter surfaces a clear `PythonInterpreterUnavailable` at preflight rather than silently falling back to an unrelated `python3`. Added a resolver→preflight regression test.
- **Windows `DEFAULT_PYTHON` → `python.exe`** — `find_on_path` stats `PATH_entry/<name>`, so the bare fallback must name `python.exe` on Windows. Added a `#[cfg(windows)]` regression test.
- **Regenerated FabricError reference** — `enum-fabricerror.mdx` now documents `PythonInterpreterUnavailable` (fields `path`/`origin`/`reason`) in place of the removed `InvalidAdapterPython`. Hand-edited surgically rather than running the generator, since the committed docs pin Rust 1.94.0 std URLs and my local toolchain is 1.95.0 — a full regen would rewrite every page's URLs.

**Skipped (with reason)**
- **Executing the interpreter as a preflight probe** — this adds a subprocess spawn to every run and still can't catch the actual failure mode (the adapter module import is inherently a runtime concern; a `python -c "pass"` probe wouldn't surface `ModuleNotFoundError` for the adapter). Existence + PATH resolution is the minimal, targeted validation for "wrong/missing interpreter."

**No change needed**
- **Python binding parity** — `to_py_error` maps `FabricError` via `.to_string()` (no per-variant match), so the error rename needs no binding update.

**Validation**: `cargo fmt --all -- --check`, `cargo test --workspace --locked` (56 in fabric-core), `cargo check --workspace --locked`, `git diff --check`, and Python `tests/adapters` + `tests/python` (251 passed) all green. (PyO3 build needed `PYO3_PYTHON` pointed at a ≥3.11 interpreter locally.)

AjayThorve (approved):

yczhang-nv:
@AjayThorve Thanks for the approval, let me handle the conflicts and get the PR merged then

PR #104, opened 2026-07-22
[DO NOT MERGE]feat: add streaming API POC

yczhang-nv:
#### Overview

Proof-of-concept for an NVIDIA NeMo Fabric streaming API built on **NeMo Relay-generated ATOF**, added as a self-contained `streaming-poc/` folder. It validates a v0.1 contract: `Runtime.invoke_stream()` as syntactic sugar over `invoke()` that yields **raw ATOF records**, with `RunResult` delivered out of band, available only when Relay is enabled — no Fabric-specific event normalization in v0.1. This is a **proposed** API surface, not implemented in the SDK; the POC models it in `streaming-poc/common/fabric_stream.py`.

**All four harnesses were run for real.** For each, the native SDK stream was captured *before* Relay (via a POC-only recorder) alongside Relay's ATOF from the same run, so the native→ATOF loss analysis is measured, not assumed. The two gateway harnesses (Claude, Codex) were captured on a **subscription / SSO** session forwarded by the Relay gateway — **no API key**. The single measured, rendering-relevant loss shared by both gateway harnesses is the per-delta token text (it lands only in the terminal scope), so the live stream is enough for a *structural* progress UI but not incremental text rendering.

This PR adds **no product code** — only the `streaming-poc/` folder (31 files; nothing under `adapters/`, `crates/`, or the Python SDK).

#### Details

- `streaming-poc/common/` — the experimental prototype: a loopback NDJSON ATOF listener (bounded queue + TCP backpressure; handles >512 KB gateway records), `invoke_stream()` (loopback endpoint injected at `start_runtime`; raw ATOF + out-of-band `result()`; **one active invocation per runtime**; early-exit via a cancellation-safe `aclose()` that drains through invocation completion), `run_harness.py`, a POC-only `native_recorder.py` that tees the native SDK stream before Relay, and `two_turn_isolation.py`.
- `streaming-poc/hermes/`, `deepagents/`, `claude/`, `codex/` — each with real `native-events.jsonl` + `events.atof.jsonl` from the same run and a `findings.md` (native→ATOF diff, measured loss, deltas-vs-terminal, duplicate-rendering risks, recommendation, reproduce steps). Deep Agents additionally includes `parallel-*.jsonl` capturing **concurrent sibling subagents** (two delegations in one message, overlapping scopes, interleaved namespaces).
[...]

PR #109, opened 2026-07-23
feat: add Relay-backed runtime streaming

yczhang-nv:
#### Overview

Implements streaming backed by NVIDIA NeMo Relay for NVIDIA NeMo Fabric
runtimes, promoting the streaming proof of concept from #104 into the supported
Python SDK surface.

`Runtime.invoke_stream()` exposes raw ATOF records while an invocation is running
and keeps the normalized terminal `RunResult` out of band. Streaming is **opt-in**:
callers pass `streaming=True` to `start_runtime(...)` (with NeMo Relay enabled) to
provision it, so NeMo Relay-enabled runtimes that do not stream pay no additional cost
and keep their existing telemetry shape. The streaming API change is additive — a
new optional `streaming` parameter — and introduces no new dependencies.

#### Details

- Add an SDK-owned loopback HTTP listener for NeMo Relay NDJSON/ATOF delivery with a
  bounded, byte-budgeted record queue (16 MiB total / 1 MiB per record), TCP
  backpressure, and a bounded post-turn drain window.
- Gate provisioning behind `start_runtime(..., streaming=True)`: a NeMo Relay-enabled
  runtime binds the loopback listener and injects the `nemo-fabric-stream` ATOF
  sink only when streaming is requested. Requesting `streaming=True` without NeMo Relay
  enabled raises `FabricConfigError`.
- Add `Runtime.invoke_stream()`, `Runtime.supports_streaming`,
  `InvokeStream.result()`, and explicit early-exit cleanup via
  `InvokeStream.aclose()`.
- Configure the listener host through an environment variable so deployments can
  override the loopback default when needed.
- Correlate records to a single invocation by ATOF scope tree: Deep Agents and
  Hermes stamp a request ID on the turn's root scope; gateway harnesses match on
  NeMo Relay's turn role and index. The Hermes correlation scope is added only for
  streaming runs, so non-streaming NeMo Relay runs keep their prior trajectory shape.
- Surface otherwise-silent failures as `RuntimeWarning`s: no NeMo Relay connection
  reached the listener, data arrived but no record matched the active turn, or a
  chunked upload was interrupted mid-turn. A clean stream completion does not warn.
- Validate streaming requests before returning an `InvokeStream`, make stream
  iteration cancellation-safe, and allow a manual `Runtime.stop()` to finalize a
  completed stream after `result()`.
- Document the public streaming contract, terminal failure semantics, adapter
  behavior, runnable examples, and the generated Python API reference.

#### Validation

- Focused streaming, adapter, generated-reference, and Fern-sync tests: 135 passed.
- `uv run --no-sync pre-commit run --all-files --show-diff-on-failure` — all hooks
  passed, including Cargo/uv lockfile freshness, attribution generation, and the
  dependency license diff.
- `just docs` — passed; Fern reported only the expected unauthenticated redirect
  check warning.
- `just test-python` — 455 passed and 44 skipped locally. Two failures and two
  setup errors remain in `test_installed_adapter_discovery.py` because uv's local
  macOS Python 3.13 virtual environments cannot load `libpython3.13.dylib`; the
  streaming-focused tests pass independently.
- End-to-end streaming turns run under
  `warnings.simplefilter("error", RuntimeWarning)` and assert non-empty records:
  - Deep Agents — `tests/e2e/test_deepagents.py` (persistent host, mock model).
  - Claude and Codex gateways — live tests in `tests/e2e/test_claude.py` and
    `tests/e2e/test_codex.py`, validating gateway turn correlation.

Start with `python/src/nemo_fabric/streaming.py` for the listener, turn
correlation, and stream lifecycle; then `python/src/nemo_fabric/client.py` for the
`streaming=True` opt-in gate and `python/src/nemo_fabric/runtime.py` for eager
validation and manual shutdown. The focused behavioral coverage is in
`tests/python/test_streaming.py`, with end-to-end coverage under `tests/e2e/`.

- Supersedes the implementation POC in #104

## Summary by CodeRabbit

* **New Features**
  * Added NeMo Relay-backed streaming to the Python SDK via `Runtime.invoke_stream()` with an `InvokeStream` async iterator of raw ATOF records and a terminal `await stream.result()`.
  * Added `Runtime.supports_streaming` and streaming-aware runtime startup/stop lifecycle handling.
  * Introduced typed v2 NeMo Relay ATOF sink configuration via file/stream sink models.

* **Documentation**
  * Added the Streaming API reference and updated Python and integration guides with end-to-end streaming workflows.
  * Updated NeMo Relay ATOF examples to use the typed v2 `sinks` model.

* **Tests**
  * Expanded/updated streaming, documentation contract, and adapter configuration tests to match the new streaming and typed sink behavior.


docs/sdk/python.mdx
```diff
@@ -314,6 +335,94 @@ on one live thread, Deep Agents invokes one compiled graph and checkpointer,
 Hermes Agent reuses one agent and session database, and Claude keeps one connected SDK
 client. Harness-native identifiers remain adapter-internal.
 
+## Relay-Backed Streaming
+
+Enable Relay before starting a runtime to consume raw ATOF records while a turn
+runs. `Runtime.supports_streaming` reports whether this Relay-backed path is
+available. It is separate from `RuntimeCapabilities.streaming`, which describes
+adapter-native progressive output.
```
AnuradhaKaruppiah: Could we make the intentional scope decision explicit here? `Runtime.invoke_stream()` exposes Relay-generated ATOF records only; adapter-native progressive output—for example, Codex app-server message and item deltas—is intentionally deferred to a future normalized Fabric streaming contract. The current text distinguishes the two capabilities, but it does not say that the deferral is deliberate.

python/src/nemo_fabric/streaming.py
```diff
@@ -0,0 +1,417 @@
...
+
+        self._closed = True
+        await self._finalize()
+
+    async def _finalize(self, *, warn_if_unconnected: bool = False) -> None:
+        if self._finalized:
+            return
+        invocation_completed = False
+        try:
+            await asyncio.shield(self._task)
```
AnuradhaKaruppiah, line 209: `_finalize()` waits for the invocation task before it starts draining the bounded record queue. If a consumer stops after the queue fills, `_emit()` blocks the NDJSON upload, so completion and cleanup depend on Relay timing out or dropping records. Could we drain concurrently while awaiting the invocation (or drain first), and add a slow-consumer test that sends more than `maxsize` records and calls `aclose()`?

python/src/nemo_fabric/streaming.py
```diff
@@ -0,0 +1,417 @@
...
+        queue = self._listener.records
+        while True:
+            if self._closed:
+                await self._finalize()
+                raise StopAsyncIteration
+            if not queue.empty():
+                return queue.get_nowait()
+            if self._task.done():
+                try:
+                    return await asyncio.wait_for(queue.get(), _DRAIN_SECONDS)
```
AnuradhaKaruppiah, line 154: A fixed 250 ms quiet window does not establish a deterministic turn boundary for this runtime-scoped listener. A late Relay record can be dropped after the previous stream finalizes or, once the next stream begins accepting, be yielded as part of the next invocation. Since `InvokeStream` promises records for one invocation, can we correlate or filter records by invocation/turn, or use an explicit Relay flush/end acknowledgement instead of best-effort timing?

AnuradhaKaruppiah (commented):
Four focused review notes: one documentation clarification about the intentionally deferred native harness-output contract, plus three robustness questions around backpressure, turn isolation, and memory bounds.

python/src/nemo_fabric/streaming.py
```diff
@@ -0,0 +1,417 @@
...
+    def __init__(
+        self,
+        *,
+        host: str = "127.0.0.1",
+        port: int = 0,
+        maxsize: int = _QUEUE_MAXSIZE,
+    ) -> None:
+        self._host = host
+        self._port = port
+        self._queue: asyncio.Queue[dict[str, Any]] = asyncio.Queue(maxsize=maxsize)
```
AnuradhaKaruppiah: `maxsize=1024` bounds record count but not memory. `_feed()` can buffer an arbitrarily large unterminated line, and the tests intentionally accept a 600 KiB record; 1,024 similarly sized parsed records would exceed 600 MiB before Python object overhead. Could we enforce a maximum record/body size or use a byte-based queue budget?

docs/sdk/python.mdx
yczhang-nv: Fixed in 1ee3e85

python/src/nemo_fabric/streaming.py
yczhang-nv: Fixed in f54b11c: _finalize() drains the queue concurrently while the invocation completes

python/src/nemo_fabric/streaming.py
yczhang-nv: Fixed in 1ee3e85: the listener enforces a 1 MiB record limit and a 16 MiB aggregate queue limit.

python/src/nemo_fabric/streaming.py
yczhang-nv: Fixed in 3508da7. The listener correlates records by Fabric request ID or Relay turn index and yields only the matched scope tree; the 250 ms window is now only a drain period.

AnuradhaKaruppiah (approved):
This looks good overall to merge for the first rel.

The entire streaming implementation is in the Python SDK which will become an issue when we publish a TS SDK or have consumers that want to directly interface with the rust API. A followup refactor is needed to set streaming up in the core in a language neutral fashion.

adapters/hermes/src/nemo_fabric_adapters/hermes/adapter.py
```diff
@@ -59,20 +83,20 @@ def build_hermes_config(
     blocked_toolsets = disabled_toolsets(payload)
 
     config: dict[str, Any] = {
-        "model": common_utils.without_none(
```
AnuradhaKaruppiah, line 62: what is the reason for not using common_utils.without_none

docs/about-nemo-fabric/overview.mdx
```diff
@@ -119,14 +119,15 @@ Harness installation and credential requirements differ by adapter. The
 contains the complete Hermes Agent environment recipe.
 
 Refer to the [Python SDK guide](/nemo/fabric/sdk/python-sdk) for planning,
-diagnostics, typed requests, and multi-turn runtime examples.
+diagnostics, typed requests, multi-turn runtimes, and Relay-backed streaming.
```
AnuradhaKaruppiah:
```suggestion
diagnostics, typed requests, multi-turn runtimes, and NeMo Relay-backed streaming.
```

docs/about-nemo-fabric/overview.mdx
```diff
@@ -119,14 +119,15 @@ Harness installation and credential requirements differ by adapter. The
...
 
 ## Choose Your Interface
 
 | Interface | Use it when | Start with |
 | --- | --- | --- |
 | Python SDK | Your application owns job config, runtime lifecycle, or multi-turn state | [Client API](/reference/api/python-library-reference/client) |
 | Runtime API | You need multiple ordered turns over one live harness runtime | [Runtime](/reference/api/python-library-reference/runtime) |
+| Streaming API | You need live Relay-generated ATOF records during a runtime turn | [Streaming](/reference/api/python-library-reference/streaming) |
```
AnuradhaKaruppiah: Update all Relay references to NeMo Relay. Use NVIDIA NeMo Relay when reviewing.

docs/sdk/python.mdx
```diff
@@ -270,8 +270,9 @@ multiple independent runtimes.
 | `Fabric.plan(config, base_dir=...)` | No | You need to inspect the selected adapter, capability mapping, and runtime capabilities before running. | Does not start a runtime. |
 | `Fabric.doctor(config, base_dir=...)` | Yes | You need preflight diagnostics for adapter availability, config support, and environment assumptions. | Checks may touch runtime dependencies. |
 | `Fabric.run(config, base_dir=..., input=...)` | Yes | You need one complete start, invoke, result, stop lifecycle. | `base_dir` is optional. Pass a `RunRequest` instead when the invocation needs IDs, context, or overrides. |
-| `Fabric.start_runtime(config, ...)` | Yes | You need state across multiple ordered invocations. | Returns a `Runtime`. Use it as an async context manager. |
+| `Fabric.start_runtime(config, ..., streaming=False)` | Yes | You need state across multiple ordered invocations. | Returns a `Runtime`. Use `streaming=True` to provision Relay-backed streaming. |
```
AnuradhaKaruppiah: Same comments for this document. Use NVIDIA NeMo Relay first time and NeMo Reelay after.

AnuradhaKaruppiah:
Have your agent scrub all references to Relay. You need to used NVIDIA NeMo Relay first use and NeMo Relay after. Double check with @dagardner-nv on docs usage.

docs/reference/api/python-library-reference/nemo_fabric.streaming.md
```diff
@@ -0,0 +1,52 @@
+---
+title: "Streaming"
+slug: "/reference/api/python-library-reference/streaming"
+description: "Consume Relay-backed raw ATOF records and terminal invocation results."
+---
+<!-- SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0 -->
+
+# <kbd>module</kbd> `nemo_fabric.streaming`
+Relay-backed streaming support for the NVIDIA NeMo Fabric Python SDK.
```
AnuradhaKaruppiah: I dont see the need for the hyphen in the Relay-backed.

docs/sdk/python.mdx
```diff
@@ -334,6 +335,114 @@ on one live thread, Deep Agents invokes one compiled graph and checkpointer,
...
+`Runtime.stop()` after the result completes finalizes the stream and discards
+those unread records.
+
+Streaming has the following v0.1 constraints:
+
+- Enable Relay and pass `streaming=True` to `start_runtime(...)`. The SDK then
+  binds its loopback listener and injects the reserved ATOF stream sink.
+  Without this opt-in, Relay configuration remains unchanged and
+  `invoke_stream(...)` raises `FabricCapabilityError`.
+- The default `streaming=False` does not force-enable ATOF for Relay runtimes
```
AnuradhaKaruppiah: I am not sure I understand this bullet. To enable streaming the user has to:
- pass "streaming=True" in start_runtime
- enable relay settings in FabricConfig 
The streaming=False description seems to imply that "ATOF for relay" is implcitly enabled if streaming is True

docs/sdk/python.mdx
```diff
@@ -334,6 +335,114 @@ on one live thread, Deep Agents invokes one compiled graph and checkpointer,
...
+- The SDK limits its queue to 1,024 records and 16 MiB of encoded record data.
+  It rejects individual records larger than 1 MiB and applies TCP backpressure.
+  Relay can drop records if a consumer stalls longer than Relay's delivery
+  timeout.
+- ATOF granularity depends on the Relay integration. Gateway harnesses expose
+  per-delta event structure, but current ATOF records keep token text in the
+  terminal scope. In-process harnesses expose scope-level progress.
+- Reconstruct nested and parallel work with `uuid` and `parent_uuid`. Stream
+  order alone does not define the Deep Agents scope tree.
+- The listener correlates records to one Relay scope tree. Deep Agents and
```
AnuradhaKaruppiah: Pls run this by @AjayThorve

docs/sdk/python.mdx
```diff
@@ -334,6 +335,114 @@ on one live thread, Deep Agents invokes one compiled graph and checkpointer,
...
+  Hermes roots carry the Fabric request ID. For Claude and Codex, the SDK
+  matches Relay turn scopes by their role and 1-based turn index. Records
+  outside the matched root and its descendants are discarded, so delayed
+  records from another turn do not enter the active stream. If the gateway
+  turn sequence does not align with the Fabric invocation sequence, the SDK
+  yields no uncorrelated records and emits a `RuntimeWarning` after natural
+  stream exhaustion. A short drain window only collects late records from the
+  matched tree; it does not define turn ownership.
+- The SDK listener binds to `127.0.0.1`. Claude and Codex gateway processes
```
AnuradhaKaruppiah: Should we make this an env variable and default to 127.0.0.1?

adapters/hermes/src/nemo_fabric_adapters/hermes/adapter.py
yczhang-nv: Restored the helper and updated the Hermes adapter to use it

docs/sdk/python.mdx
yczhang-nv: Updated the doc so that it clarified that `streaming=False` leaves the existing Relay configuration unchanged

docs/sdk/python.mdx
yczhang-nv: The SDK listener now reads `NEMO_FABRIC_STREAMING_HOST`, defaulting to `127.0.0.1`. I also updated the documentation to note that access to a non-loopback interface should be restricted. Implemented in 632e809.

python/src/nemo_fabric/streaming.py
```diff
@@ -0,0 +1,647 @@
...
+            if not queue.empty():
+                return queue.get_nowait()
+            if self._task.done():
+                try:
+                    return await asyncio.wait_for(queue.get(), _DRAIN_SECONDS)
+                except TimeoutError:
+                    await self._finalize(warn_if_unavailable=True)
+                    raise StopAsyncIteration from None
+
+            getter = asyncio.create_task(queue.get())
```
AjayThorve, line 159: Cancellation can leave an orphaned queue consumer here. If the task executing `__anext__()` is cancelled while blocked in `asyncio.wait()`, the separately created `getter` task remains alive because `asyncio.wait()` does not cancel it.

```text
Expected:

cancelled __anext__ -> internal getter cancelled
Relay -> queue -> next __anext__ receives "first"

Current:

cancelled __anext__ -> orphaned getter remains
Relay -> queue -> orphaned getter consumes "first" -> no caller
next __anext__ receives "second"
```

This is a single-consumer queue; the cancelled call effectively leaves behind a ghost second consumer that silently steals the next record. This can occur naturally when an HTTP streaming request disconnects, times out, or is cancelled during shutdown.

Please make the getter cancellation-safe and add a regression test that cancels a pending `__anext__()`, emits two records, and verifies the next call still receives the first. The cleanup also needs to handle the race where the getter completes concurrently with cancellation—if it already consumed a record, that record must be retained for the next `__anext__()` rather than discarded.

yczhang-nv: Thanks @AjayThorve ! Fixed in 6b7e0dd. Cancellation now cleans up the internal getter, while records consumed during a cancellation race are retained for the next __anext__() call. Added regression tests for both cases.

AjayThorve (approved):
This looks good to me after the cleanup! Great job

PR #154, opened 2026-07-29
fix: rewrite version pins with dependency extras

AnuradhaKaruppiah:
## Summary

- Rewrite exact internal dependency pins that include extras, such as `nemo-fabric-adapters-hermes[harness]`.
- Move the Python version rewrite into a testable script and add regression coverage for extras and environment markers.

## Root cause

The release regex matched bare package names only, so RC builds updated package versions but left root extras pinned to the stable version. Installing a root extra could therefore request an adapter version that was never published.

## Validation

- `just --set no_uv true test-python` — 531 passed, 14 skipped
- `cargo check -p fabric-python --locked`
- `just wheels`
- Simulated `v0.2.0-rc5` stamping and verified the built root wheel pins all adapter extras to `0.2.0rc5`

## Summary by CodeRabbit

* **Chores**
  * Improved the Python package release version update workflow by switching to a dedicated CLI utility.
  * Automatically synchronizes versions and internal dependency pins across the main project and adapter packages.
  * Enforces that the runtime project keeps dynamic versioning and validates version/pin consistency.
* **Tests**
  * Added coverage ensuring version and internal pin updates apply to adapter projects, including optional dependency extras and environment markers.

### DIFF
--- a/justfile
+++ b/justfile
@@ -245,58 +245,7 @@ set_python_project_versions() {
     local python_executable=""
     version="$(semver_to_pep440 "$1")"
     python_executable="$(uv_python_executable)"
-
-    "$python_executable" - "$version" <<'PY'
-from pathlib import Path
-import re
-import sys
-import tomllib
-
-version = sys.argv[1]
-project_paths = (
-    Path("pyproject.toml"),
-    *sorted(Path("adapters").glob("**/pyproject.toml")),
-)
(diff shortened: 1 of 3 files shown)


dagardner-nv (approved):

zhongxuanwang-nv (approved):
Thanks Anuradha!!!

PR #175, opened 2026-08-05
feat: add external NAT reference adapter

AjayThorve:
#### Overview

Adds a source-only NeMo Agent Toolkit (NAT) adapter under `external/` as the first third-party reference implementation of the public NeMo Fabric adapter contract.

The adapter translates typed `FabricConfig` input into an in-memory NAT `Config`, loads installed NAT components, and owns one persistent `WorkflowBuilder` and `SessionManager` across `start`, repeated `invoke`, and `stop`. This PR intentionally does not add wheel metadata, bundled catalog entries, or installed-adapter discovery wiring.

The typed workflow contract is already available on `main` through #176. This PR now contains only the external NAT reference adapter, its examples, and focused tests.

#### Details

- Adds the `nvidia.fabric.nat` descriptor and generic Python lifecycle runner.
- Selects NAT workflows through root `workflow.entrypoint` and maps `workflow.settings` into the selected NAT component; no NAT workflow YAML or `config_file` path is used.
- Uses descriptor `workflow_schema` as the single workflow support and validation claim. `harness.settings` is limited to registered NAT `functions` and `function_groups`.
- Maps normalized models and ReAct system instructions. Built-in ReAct behavior is applied only to NAT's short and canonical qualified ReAct references, so a third-party `custom/react_agent` is not mutated.
- Maps per-server MCP `allowed_tools` and `blocked_tools` into NAT `include` or `exclude` policy. An explicit empty allowlist fails closed by omitting the generated group because NAT interprets `include=[]` as unfiltered.
- Applies root `tools.enabled` and `tools.blocked` across NAT functions, function groups, and `<group>__<member>` selectors.
- Loads custom workflows, functions, and function groups through installed NAT `nat.components` registrations.
- Adds typed calculator-MCP and NAT-native email-phishing examples using the same root workflow contract. The calculator example includes a source-only stdio MCP server and requires no separately managed endpoint.
- Normalizes errors without exposing underlying exception text in NeMo Fabric results or stderr artifacts.

#### Validation

- `uv run --no-sync pytest -q tests/adapters/test_external_nat_adapter.py` — 52 passed, 2 optional real-NAT tests skipped when NAT is not installed.
- `just test-python` — 652 passed, 17 skipped.
- A direct MCP stdio protocol check discovered `add`, `subtract`, `multiply`, and `divide`, and returned `42.0` for `multiply(21, 2)`.
- Real NAT 1.8 typed-`Config` validation passed for both `react_agent` and `nat.plugins.langchain.agent.react_agent/react_agent`.
- Both typed examples produce current NeMo Fabric plans against the external descriptor.
- Changed-file pre-commit hooks and `git diff --check` passed.

Live model invocation was not exercised because it requires credentials.

Start with `_nat_workflow`, `_is_react_agent`, and `build_nat_config_mapping` in `external/nat/src/nemo_fabric_adapters/nat/adapter.py`. Then review `workflow_schema` in `external/nat/fabric-adapter.json` and the root-workflow/example coverage in `tests/adapters/test_external_nat_adapter.py`.

- Relates to FABRIC-117
- Relates to #176

### DIFF
--- a/external/nat/README.md
+++ b/external/nat/README.md
@@ -21,7 +21,7 @@ that have no portable NeMo Fabric equivalent.
 
 | NeMo Fabric input | NAT configuration |
 | --- | --- |
-| `models.<role>` | `llms.<role>`; every Fabric model-role name is preserved |
+| `models.<role>` | `llms.<role>`; every NeMo Fabric model-role name is preserved |
 | `instructions.system` | Built-in `react_agent` workflow `additional_instructions`; other workflow types reject this field in the initial adapter |
 | `workflow.entrypoint.kind=nat_workflow` | Resolve a registered NAT workflow component |
 | `workflow.entrypoint.ref` | `workflow._type` |
@@ -47,19 +47,19 @@ streaming, or live-update support.
 ## MCP Tool Filters
 
 The adapter consumes the routed `capability_plan.native.mcp_servers` entries,
-including the normalized per-server filters. Fabric MCP tool names remain bare
+including the normalized per-server filters. NeMo Fabric MCP tool names remain bare
 server-local names; NAT exposes a selected member as `<server>__<tool>`.
 
-| Fabric server policy | Generated NAT function group |
+| NeMo Fabric server policy | Generated NAT function group |
 | --- | --- |
 | `allowed_tools` omitted and `blocked_tools=[]` | No `include` or `exclude`; expose all discovered tools |
 | Nonempty `allowed_tools` only | `include=allowed_tools` |
 | `allowed_tools` omitted and nonempty `blocked_tools` | `exclude=blocked_tools` |
-| Both lists configured | `include=allowed_tools`; Fabric requires `blocked_tools` to be disjoint, so those names are already outside the allowlist |
+| Both lists configured | `include=allowed_tools`; NeMo Fabric requires `blocked_tools` to be disjoint, so those names are already outside the allowlist |
 | `allowed_tools=[]` | Omit the generated group; expose no tools from that server |
 
 NAT rejects a function group that sets both `include` and `exclude`, so the
-adapter emits only `include` whenever an allowlist is present. Fabric rejects
+adapter emits only `include` whenever an allowlist is present. NeMo Fabric rejects
 blank names and an allow/block overlap before adapter startup. A nonempty
 generated MCP group is added to workflows that expose `tool_names`; callers do
 not repeat portable MCP servers in `harness.settings`. A workflow implementation
@@ -95,11 +95,10 @@ cp external/nat/fabric-adapter.json \
   .tmp/nat-reference/adapters/nat/fabric-adapter.json
 ```
 
-The calculator example also requires a streamable-HTTP calculator MCP server.
-Set its endpoint and run the typed `FabricConfig` example:
+The calculator example starts its source-only MCP server over stdio, so it does
+not require a separately managed endpoint. Run the typed `FabricConfig` example:
 
 ```bash
-export CALCULATOR_MCP_URL=http://127.0.0.1:9901/mcp
 uv run python external/nat/examples/calculator.py \
   --base-dir "$PWD/.tmp/nat-reference"
 ```
--- a/external/nat/README.md
+++ b/external/nat/README.md
@@ -15,14 +15,17 @@ calculator or email-phishing components.
 
 ## Configuration Boundary
 
-NeMo Fabric owns portable configuration. `harness.settings` contains only
-NAT-native component configuration that has no portable NeMo Fabric equivalent.
+NeMo Fabric owns portable configuration. `workflow` selects and configures the
+NAT executable, while `harness.settings` contains NAT-native function components
+that have no portable NeMo Fabric equivalent.
 
 | NeMo Fabric input | NAT configuration |
 | --- | --- |
 | `models.<role>` | `llms.<role>`; every Fabric model-role name is preserved |
 | `instructions.system` | Built-in `react_agent` workflow `additional_instructions`; other workflow types reject this field in the initial adapter |
-| `harness.settings.workflow` | `workflow` |
+| `workflow.entrypoint.kind=nat_workflow` | Resolve a registered NAT workflow component |
+| `workflow.entrypoint.ref` | `workflow._type` |
+| `workflow.settings` | Remaining `workflow` component fields |
 | `harness.settings.functions` | `functions` |
 | `harness.settings.function_groups` | `function_groups` |
 | Harness-native `mcp.servers.<name>` | Generated `mcp_client` function group named `<name>` |
@@ -31,7 +34,8 @@ NAT-native component configuration that has no portable NeMo Fabric equivalent.
 The adapter loads installed `nat.components` entry points before NAT validates
 the generated configuration. A custom function, function group, or workflow is
 therefore supplied as an installed NAT component package and selected by its
-registered `_type`. No Python import path or callable crosses `FabricConfig`.
+registered type in `workflow.entrypoint.ref` or the component `_type` in
+`harness.settings`. No Python import path or callable crosses `FabricConfig`.
 
 At runtime, `start` loads components, enters one `WorkflowBuilder`, creates a
 `SessionManager` with that shared builder, and retains both resources. Each
--- a/external/nat/README.md
+++ b/external/nat/README.md
@@ -106,7 +106,7 @@ run the example:
 
 ```bash
 uv pip install -e \
-  /path/to/NeMo-Agent-Toolkit/examples/evaluation_and_profiling/email_phishing_analyzer
+  "<path-to-nat-checkout>/examples/evaluation_and_profiling/email_phishing_analyzer"
 uv run python external/nat/examples/email_phishing.py \
   --base-dir "$PWD/.tmp/nat-reference"
 ```
--- a/external/nat/README.md
+++ b/external/nat/README.md
@@ -1,0 +1,115 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# NVIDIA NeMo Fabric NAT Reference Adapter
+
+This source-only adapter runs an NVIDIA NeMo Agent Toolkit (NAT) workflow behind
+the NeMo Fabric lifecycle contract. It is a third-party adapter reference, not a
+bundled NeMo Fabric adapter or a published package.
+
+The implementation is generic. It constructs NAT configuration in memory from
+`FabricConfig`; it does not read a NAT YAML file and does not hardcode the
+calculator or email-phishing components.
+
+## Configuration Boundary
+
+NeMo Fabric owns portable configuration. `harness.settings` contains only
+NAT-native component configuration that has no portable NeMo Fabric equivalent.
+
+| NeMo Fabric input | NAT configuration |
+| --- | --- |
+| `models.<role>` | `llms.<role>`; every Fabric model-role name is preserved |
+| `instructions.system` | Built-in `react_agent` workflow `additional_instructions`; other workflow types reject this field in the initial adapter |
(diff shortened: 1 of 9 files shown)


external/nat/examples/calculator.py
```diff
@@ -0,0 +1,86 @@
...
+        metadata=MetadataConfig(
+            name="nat-calculator",
+            description="Uses calculator tools exposed by an MCP server.",
+        ),
+        harness=HarnessConfig(
+            adapter_id="nvidia.fabric.nat",
+            resolution="preinstalled",
+            settings={
+                "workflow": {
+                    "_type": "react_agent",
```
AnuradhaKaruppiah: does the per_user_react_agent vs react_agent matter here?

external/nat/examples/calculator.py
```diff
@@ -0,0 +1,86 @@
...
+            system=InstructionConfig(
+                content="Use the calculator tools for arithmetic. Return a concise answer."
+            )
+        ),
+        runtime=RuntimeConfig(input_schema="text", output_schema="message"),
+    )
+    config.add_mcp_server(
+        "calculator",
+        transport="streamable-http",
+        url=os.environ.get("CALCULATOR_MCP_URL", "http://127.0.0.1:9901/mcp"),
```
AnuradhaKaruppiah: should we use stdio instead to make testing easier

external/nat/examples/email_phishing.py
```diff
@@ -0,0 +1,89 @@
...
+            settings={
+                "functions": {
+                    "email_phishing_analyzer": {
+                        "_type": "email_phishing_analyzer",
+                        "llm": "default",
+                    }
+                },
+                "workflow": {
+                    "_type": "react_agent",
+                    "llm_name": "default",
```
AnuradhaKaruppiah: yup, this is going to be a pain to normalize. we need to find a way to collapse config vs. workflow reference in a somewhat harness agnostic way.

AnuradhaKaruppiah (commented):
Partial review only (read the PR to get an idea of the examples)

external/nat/examples/calculator.py
AjayThorve: it does matter actually. `per_user_react_agent` invocation needs a `user_id` i think, so it would just fail for the current implementation (`request.context.user_id`). 

I think the adapter would need some updating, as per user workflows have a slightly different system instruction and tool mapping as well, but from a quick glance, should be straight-forward. 

Great catch!

external/nat/fabric-adapter.json
```diff
@@ -0,0 +1,73 @@
...
+    "accepts": [
+      "models",
+      "models.base_url",
+      "models.temperature",
+      "instructions.system",
+      "tools.enabled",
+      "tools.blocked",
+      "mcp",
+      "mcp.tool_filters"
+    ]
```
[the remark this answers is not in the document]
AjayThorve: Not applying this suggestion because it conflicts with the current typed descriptor contract. `AdapterConfigField` serializes the supported values as `models`, `models.base_url`, `models.temperature`, `instructions.system`, `tools.enabled`, `tools.blocked`, `mcp`, and `mcp.tool_filters`; `instructions` and `tools` are not valid variants and would fail descriptor deserialization. The compatibility and capability planners check these exact variants, and `test_descriptor_declares_exact_source_reference_contract` freezes the list.

external/nat/examples/calculator.py
AjayThorve: The reference example intentionally selects `react_agent`. This adapter does not advertise `per_user_react_agent` support, so the example does not require `request.context.user_id`. Per-user workflow semantics can be added separately when they are mapped and tested end to end.

external/nat/examples/calculator.py
AjayThorve: Addressed in `16c360c3`. The calculator example now launches a source-only calculator MCP server over stdio and no longer requires a separately managed HTTP endpoint. The MCP protocol path was also exercised directly.

external/nat/examples/email_phishing.py
AjayThorve: Addressed by the typed workflow contract merged in #176 and adopted here. Workflow selection now uses root `workflow.entrypoint`, while immutable component construction values use `workflow.settings`; the adapter derives NAT `_type` from the entrypoint reference. This thread became outdated after the rebase.

tests/python/test_workflow_validation.py
```diff
@@ -0,0 +1,137 @@
...
+            resolution="preinstalled",
+        ),
+        workflow=_workflow(llm_name="default"),
+    )
+
+    with pytest.raises(
+        FabricConfigError,
+        match="does not declare a workflow_schema",
+    ):
+        Fabric().plan(config, base_dir=tmp_path)
```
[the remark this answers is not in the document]
AjayThorve: This thread belongs to the workflow-contract changes merged independently in #176. After rebasing, PR #175 no longer changes `tests/python/test_workflow_validation.py`; core workflow validation follow-ups are outside this NAT adapter PR.

tests/python/test_workflow_validation.py
```diff
@@ -0,0 +1,137 @@
...
+
+def test_workflow_schema_controls_whether_workflow_is_required(tmp_path: Path):
+    _write_descriptor(tmp_path)
+
+    with pytest.raises(FabricConfigError, match=r"at `workflow`"):
+        Fabric().plan(_config(), base_dir=tmp_path)
+
+    _write_descriptor(tmp_path, optional=True)
+    plan = Fabric().plan(_config(), base_dir=tmp_path)
+    assert plan.config.workflow is None
```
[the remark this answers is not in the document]
AjayThorve: This thread belongs to the workflow-contract changes merged independently in #176. After rebasing, PR #175 no longer changes `tests/python/test_workflow_validation.py`; this test refactor is outside the NAT adapter PR.

external/nat/src/nemo_fabric_adapters/nat/adapter.py
```diff
@@ -0,0 +1,915 @@
...
+    if transport in {"http", "streamablehttp"}:
+        transport = "streamable-http"
+    if transport not in {"sse", "streamable-http"}:
+        raise _config_error(
+            "nat_unsupported_mcp_transport",
+            f"NAT MCP server {name!r} has unsupported transport {transport!r}",
+            server=name,
+            transport=transport,
+        )
+    return {"transport": transport, "url": target}
```
zhongxuanwang-nv, line 356: Should we also include stdio MCP's `args` and `env`?

external/nat/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+| NeMo Fabric input | NAT configuration |
+| --- | --- |
+| `models.<role>` | `llms.<role>`; every NeMo Fabric model-role name is preserved |
+| `instructions.system` | Built-in `react_agent` workflow `additional_instructions`; other workflow types reject this field in the initial adapter |
+| `workflow.entrypoint.kind=nat_workflow` | Resolve a registered NAT workflow component |
+| `workflow.entrypoint.ref` | `workflow._type` |
+| `workflow.settings` | Remaining `workflow` component fields |
+| `harness.settings.functions` | `functions` |
+| `harness.settings.function_groups` | `function_groups` |
```
zhongxuanwang-nv, line 30: i'm just curious, would `function_groups` be something too NAT-specific? Should we still bring it to the overarching `harness.settings`?

zhongxuanwang-nv (approved):
Thanks Ajay!! For the `env` and `arg` thing, I also opened another [PR -178](https://github.com/NVIDIA/NeMo-Fabric/pull/178) to address this if you also think this is an issue, I think this PR is mergeable imo

external/nat/README.md
AnuradhaKaruppiah: I agree, functions and function_groups need to become tools. type information about them should be conveyed via ToolConfig.extensions

external/nat/examples/email_phishing.py
```diff
@@ -0,0 +1,95 @@
...
+                api_key_env="NVIDIA_API_KEY",
+                temperature=0.0,
+            )
+        },
+        instructions=InstructionsConfig(
+            system=InstructionConfig(
+                content='State whether the email is "phishing" or "benign" and explain why.'
+            )
+        ),
+        tools=ToolsConfig(enabled=["email_phishing_analyzer"]),
```
AnuradhaKaruppiah, line 69: I see what is happening here! we do need a way to actually define a tool before referencing.

We can introduce constructs that match functions and function groups (they don't need to be called that) in the normalized FabricConfig and then reference them as tools. In today's harness tool customization is via MCPs but that mandating that every tool be represented as a stdio-MCP is not performant. @zhongxuanwang-nv what is the LG version of tools?

external/nat/src/nemo_fabric_adapters/nat/adapter.py
```diff
@@ -0,0 +1,915 @@
...
+            from nat.builder.workflow_builder import WorkflowBuilder
+            from nat.runtime.session import SessionManager
+
+            config = build_nat_config(payload)
+            builder = await stack.enter_async_context(
+                WorkflowBuilder.from_config(config=config)
+            )
+            sessions = await SessionManager.create(
+                config=config,
+                shared_builder=builder,
```
AnuradhaKaruppiah, line 816: The shared builder is throwing me off a little bit as from the top level we expect the workflow to be per-user-react-agent. but internally there is no value in exercising the per-user path as we anyway setup a separate builder for each user, right?

AjayThorve: Yes, that is the intended behavior. `shared_builder` here is the runtime-level builder containing the shared NAT configuration and registered components; it does not mean that every user shares the same workflow instance.

`SessionManager.create(...)` inspects the selected NAT workflow registration:

 - For `react_agent`, it builds one shared workflow from the builder.
 - For `per_user_react_agent`, it retains the shared builder but creates and caches a separate `PerUserWorkflowBuilder` and workflow for each `user_id` passed to `session(...)`.

 Fabric creates one adapter runtime for the configured agent and passes the invocation’s `user_id` into NAT. NAT therefore remains responsible for the per-user workflow lifecycle and isolation; Fabric should not create a separate top-level `WorkflowBuilder` itself for every user.

 One improvement is needed in this PR: the adapter currently recognizes only `react_agent` when applying normalized system instructions and tool mappings. We should also recognize `per_user_react_agent` and its fully qualified registry reference, add mapping tests for it, and document that both shared and per-user ReAct workflows are supported. The `start`/`invoke`/`stop` lifecycle and use of `shared_builder` can otherwise remain unchanged.

external/nat/examples/email_phishing.py
AjayThorve: yep, `tools.enabled/blocked` are currently selection policy, not definitions

AjayThorve: we may need an updated `tools.definitions`, with a typed per-tool extension/schema mechanism.

external/nat/README.md
AjayThorve: `ToolConfig.extensions` currently only supports selection policy, so before updating this, we'd need to introduce `ToolConfig.extensions.definitions` or something similar, i will work on the design and we can discuss

external/nat/src/nemo_fabric_adapters/nat/adapter.py
AjayThorve: i see you already have a follow-up PR, but a pretty quick fix, instead of separating it, can you just push it to my branch, will show up in this PR itself

external/nat/src/nemo_fabric_adapters/nat/adapter.py
AjayThorve: will create a follow up task to address per-user workflow gaps

external/nat/examples/email_phishing.py
AjayThorve: will create a follow-up task for this

external/nat/README.md
AjayThorve: follow-up tasks

PR #181, opened 2026-08-06
fix: wait for Relay ATIF finalization

AjayThorve:
#### Overview

Fix the Relay artifact lifecycle race for both Codex and Claude without
upgrading Relay or changing the observability schema. Fabric now waits for the
configured local ATIF exporter to finish before it snapshots relay_artifacts,
so a successful invocation cannot be returned while its trajectory is still in
Relay's asynchronous subscriber queue.

The readiness logic is shared in adapters/common. The existing
nemo-relay>=0.6.0,<0.7 contract remains unchanged, so this fix is suitable for
the release/0.1 line and supersedes the Relay 0.7 dependency-upgrade mitigation
in #180.

This branch also sets the release version to 0.1.1 across the Rust workspace,
all seven Python distributions, exact internal dependency pins, and lockfiles.

#### Details

- Snapshot cheap filesystem fingerprints (device, inode, size, and nanosecond
  mtime) for runtime-scoped ATIF files before an invocation.
- After a successful Codex or Claude turn, inspect only new or
  metadata-changed paths and return as soon as one contains a complete JSON
  object. This supports a new file, atomic replacement, overwrite, and append.
- Do not hash, read, or parse unchanged artifacts from prior turns.
- Keep JSON parsing as the readiness condition because Relay writes directly to
  the final path; file existence or timestamps do not prove the write is
  complete. Metadata selects candidates, while successful JSON parsing proves
  readiness.
- Use a hard five-second deadline instead of a fixed sleep. A timeout returns
  codex_relay_atif_timeout or claude_relay_atif_timeout and makes the persistent
  runtime unavailable, preventing a late artifact from being attributed to a
  later turn.
- Return an empty relay_artifacts list on timeout and bypass artifact rescanning
  on the later runtime-unavailable response.
- Skip the local-file wait when ATIF uses non-empty remote storage, because
  Relay does not produce a local ATIF in that mode.
- Preserve ATOF collection, streaming, persistent SDK sessions, and the
  runtime-owned Relay gateway.

Codex and Claude need this guard because both collect artifacts immediately
after an SDK terminal result while Relay hook events are dispatched
asynchronously. Hermes already flushes subscribers before collection, and Deep
Agents collects after leaving its Relay plugin context, so those adapters are
unchanged.

Runtime-scoped output directories keep parallel evaluator sessions isolated.
The wait is condition-based and normally adds only the time Relay needs to
finish the file.

#### Release packaging

- Ran just set-version 0.1.1.
- Updated Cargo and Python package versions plus exact internal 0.1.1 pins.
- Regenerated the Python lockfiles with the CI-pinned uv 0.9.22. Dependency
  versions are unchanged; the locks add newly published Relay 0.6 source and
  musllinux artifact records.
- Regenerated Python attributions; the only generated attribution change is the
  aiohttp license-file path.
- No public configuration or API contract changed.

#### Validation

- just test-python — 582 passed, 15 skipped.
- just test-rust — 68 passed.
- cargo check --workspace --locked — passed.
- cargo fmt --all -- --check — passed.
- just build-python — passed for all seven 0.1.1 distributions.
- just wheels — built all seven wheels; wheel metadata and exact internal
  0.1.1 pins were verified.
- Pre-commit on all files — passed, including Ruff, GitHub Actions lint,
  Cargo/uv lock freshness, generated attributions, and dependency-license
  checks.
- Real Relay 0.6 Codex and Claude E2E — passed one-shot artifact semantics and
  persistent two-turn streaming runtimes.
- Platform Codex and Claude matrix at p=1,3,5,10 — 80/80 total trials completed
  with isolated runtime-scoped ATIF directories, valid ATIF-v1.7 trajectories,
  non-empty steps, and zero contract failures.
- Exact final-head Platform regression at p=10 — Codex 10/10 and Claude 10/10,
  with 20 distinct runtime-scoped ATIF directories and zero failures.
- Exact final-head non-target Platform regression at p=5 — Deep Agents 5/5 and
  Hermes 5/5 against Relay 0.6, with ten distinct runtime-scoped directories,
  valid trajectories and promoted trace evidence, and zero failures.

Documentation is unchanged because public configuration and dependency
contracts are unchanged. There are no breaking changes.

Start with
adapters/common/src/nemo_fabric_adapters/common/relay_artifacts.py, then the
calls from CodexRuntime.invoke() and ClaudeRuntime.invoke(). The focused tests
cover delayed and partial writes, unchanged-path exclusion, changed-path
detection, remote-only configuration, the hard timeout, no-rescan timeout
behavior, unusable-runtime behavior, and parallel runtime-directory isolation.

- Relates to #180 and the NeMo Platform 0.4 ATIF release blocker.

## Summary by CodeRabbit

* **New Features**
  * Added reliable Relay ATIF artifact tracking for supported Codex and Claude sessions.
  * Turns now wait for newly generated, valid artifacts before completing.
  * Added a five-second timeout for artifact finalization.

* **Bug Fixes**
  * Prevented stale or incomplete artifacts from being treated as finalized.
  * Improved isolation when multiple sessions await artifact generation.
  * Timed-out runtimes are marked unavailable to prevent subsequent use.
  * Artifact finalization timeouts now return clear, non-retryable errors.
  * Runtime failures now report a clear, generic error message.

### DIFF
--- a/adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
+++ b/adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
@@ -1078,7 +1078,7 @@ class CodexRuntime:
             _output_schema(payload)
             relay = self._relay
             atif_before = (
-                relay_artifacts.snapshot_atif_paths(relay.plugin_config)
+                relay_artifacts.snapshot_atif_files(relay.plugin_config)
                 if relay is not None
                 and relay_artifacts.expects_local_atif(relay.plugin_config)
                 else None
--- a/adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
+++ b/adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
@@ -1065,11 +1065,10 @@ class CodexRuntime:
             "request": invocation.get("request"),
         }
         if self._unusable:
-            output = _failure(
+            return _failure(
                 "codex_runtime_unavailable",
                 "Codex runtime cannot accept another invocation after a runtime failure",
             )
-            return _relay_output(output, self._relay) if self._relay else output
 
         try:
             request_prompt(payload)
--- a/adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
+++ b/adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
@@ -923,7 +923,12 @@ async def _invoke_thread(
         return sdk_failure(error), False
 
 
-def _relay_output(output: dict[str, Any], relay: CodexRelaySettings) -> dict[str, Any]:
+def _relay_output(
+    output: dict[str, Any],
+    relay: CodexRelaySettings,
+    *,
+    artifacts: list[dict[str, str]] | None = None,
+) -> dict[str, Any]:
     output["relay_runtime"] = {
         "enabled": True,
         "emitter": "codex-sdk/nemo-relay",
@@ -932,8 +937,10 @@ def _relay_output(output: dict[str, Any], relay: CodexRelaySettings) -> dict[str
         "gateway_url": relay.gateway.url,
         "gateway_log_path": str(relay.gateway.log_path),
     }
-    output["relay_artifacts"] = common_utils.collect_relay_artifacts(
-        relay.plugin_config
+    output["relay_artifacts"] = (
+        common_utils.collect_relay_artifacts(relay.plugin_config)
+        if artifacts is None
+        else artifacts
     )
     return output
 
@@ -1092,12 +1099,19 @@ class CodexRuntime:
                     relay.plugin_config, atif_before
                 )
                 if finalized is None:
-                    raise AdapterRelayError(
-                        "codex_relay_atif_timeout",
-                        "NeMo Relay did not finalize an ATIF artifact before the deadline",
-                        metadata={
-                            "timeout_seconds": relay_artifacts.ATIF_FINALIZATION_TIMEOUT_SECONDS,
-                        },
+                    self._unusable = True
+                    return _relay_output(
+                        adapter_failure(
+                            AdapterRelayError(
+                                "codex_relay_atif_timeout",
+                                "NeMo Relay did not finalize an ATIF artifact before the deadline",
+                                metadata={
+                                    "timeout_seconds": relay_artifacts.ATIF_FINALIZATION_TIMEOUT_SECONDS,
+                                },
+                            )
+                        ),
+                        relay,
+                        artifacts=[],
                     )
         except AdapterRelayError as error:
             output = adapter_failure(error)
--- a/adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
+++ b/adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 import asyncio
-import hashlib
 import json
 import logging
 import math
@@ -33,14 +32,13 @@ from openai_codex.types import Personality, ReasoningEffort, TurnStatus
 
 import nemo_fabric_adapters.common.relay_gateway as relay_gateway
 import nemo_fabric_adapters.common.relay_hooks as relay_hooks
+import nemo_fabric_adapters.common.relay_artifacts as relay_artifacts
 import nemo_fabric_adapters.common.utils as common_utils
 from nemo_fabric_adapters.common import lifecycle
 
 
 DEFAULT_TIMEOUT_SECONDS = 1800.0
 INTERRUPT_TIMEOUT_SECONDS = 5.0
-RELAY_ATIF_FINALIZATION_TIMEOUT_SECONDS = 5.0
-RELAY_ATIF_POLL_INTERVAL_SECONDS = 0.05
 SANDBOXES = {
     "read-only": Sandbox.read_only,
     "workspace-write": Sandbox.workspace_write,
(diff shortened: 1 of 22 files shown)


adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
```diff
@@ -38,6 +39,8 @@
 
 DEFAULT_TIMEOUT_SECONDS = 1800.0
 INTERRUPT_TIMEOUT_SECONDS = 5.0
+RELAY_ATIF_FINALIZATION_TIMEOUT_SECONDS = 5.0
```
willkill07: Depending on the size of the trajectory, do you think 5 sec is a reasonable upper-bound?

adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
```diff
@@ -937,6 +940,83 @@ def _relay_output(output: dict[str, Any], relay: CodexRelaySettings) -> dict[str
...
+        try:
+            contents = path.read_bytes()
+        except (OSError, RuntimeError, ValueError):
+            continue
+        if before.get(path) == hashlib.sha256(contents).digest():
+            continue
+        try:
+            document = json.loads(contents)
+        except (UnicodeDecodeError, json.JSONDecodeError):
+            continue
```
willkill07: I understand the safety checks here, but it's quite expensive to load the file, read the bytes, and parse as JSON before validating.

Isn't the existence of the file enough since you are expecting it to have a particular name?

Why not compare create time to modify time?

adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
AjayThorve: for the platform evaluator trajectories,  size ranged from 0.4–3.36 MB, parsing the largest took around 7.5ms, so this seemed like ample time. Having said that, i don't think relay has any upper bound trajectory size, so its more of an evidence based guess if anything. Open to suggestions

willkill07: No, this is fine :) I just wasn't sure.

adapters/common/src/nemo_fabric_adapters/common/relay_artifacts.py
```diff
@@ -0,0 +1,109 @@
...
+    plugin_config: dict[str, Any], before: AtifSnapshot
+) -> Path | None:
+    """Find a new or changed ATIF path containing a complete JSON object."""
+
+    current = snapshot_atif_files(plugin_config)
+    for path in sorted(current):
+        if before.get(path) == current[path]:
+            continue
+        try:
+            # Relay writes directly to the final path, so existence alone does
```
AnuradhaKaruppiah, line 82: agree with this...

adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
```diff
@@ -1066,7 +1067,7 @@ async def invoke(self, invocation: dict[str, Any]) -> dict[str, Any]:
         if self._unusable:
             output = _failure(
                 "codex_runtime_unavailable",
-                "Codex runtime cannot accept another invocation after an SDK failure",
+                "Codex runtime cannot accept another invocation after a runtime failure",
             )
             return _relay_output(output, self._relay) if self._relay else output
```
AnuradhaKaruppiah: After an ATIF timeout, the runtime becomes unusable. If the ATIF finishes late, the next invocation enters this branch and `_relay_output()` rescans the directory, potentially attaching that late artifact to the later `codex_runtime_unavailable` response. This appears to contradict the goal of preventing late artifact attribution. Could the unusable path omit artifact collection, as Claude does, and could the timeout test cover a late file arriving before the next invocation?

AnuradhaKaruppiah (approved):
LGTM. left one minor agent suggested comment.

adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
AjayThorve: Good point. I changed this so we don’t read every existing ATIF file. Before a turn, Fabric snapshots cheap metadata for the current files—device, inode, size, and nanosecond mtime. Afterward, it only opens a path if it is new or that metadata changed. That covers a new file, atomic replacement, overwrite, and append.
We still JSON-parse the changed candidate because Relay opens the final filename before the write is complete. File existence is not a completion signal, and on POSIX ctime is inode-change time—not creation time—so comparing ctime and mtime still cannot tell us that the writer has finished. A successful parse is the only completion signal available today.
I also hardened the timeout path: Fabric returns an empty artifact list and marks the runtime unusable without rescanning, so a late write cannot be attached to a later turn. Longer term, the cleaner contract is a Relay flush/completion acknowledgment or temp-file plus atomic rename.

adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
AjayThorve: Good catch—fixed in the current head. The unusable Codex path now returns directly without `_relay_output`, and the timeout path passes `artifacts=[]` instead of rescanning the Relay directory. The regression test starts with a pre-existing ATIF, lets a late ATIF arrive after the timeout and before the next invocation, and verifies that the timeout response has an empty artifact list while the subsequent `codex_runtime_unavailable` response contains no Relay artifact fields. Claude uses the same timeout guard.

dagardner-nv (approved):
Approving for deps

adapters/common/src/nemo_fabric_adapters/common/relay_artifacts.py
AjayThorve: Addressed in the current head. Fabric now snapshots cheap per-path metadata before the turn and only opens ATIF candidates that are new or changed, covering create, atomic replace, overwrite, and append without rereading old trajectories. We still parse the changed candidate because Relay writes directly to the final filename, so existence and timestamps cannot prove the write has finished. The shared tests cover the changed-path cases and parallel runtime-directory isolation.

PR #188, opened 2026-08-07
fix: reject colliding tool definition extensions

zhongxuanwang-nv:
#### Overview

Prevents `ToolsConfig.add_definition()` from silently allowing `extra_fields` to overwrite normalized `kind`, `ref`, or `settings` values.

#### Details

- Reject extension fields named `kind`, `ref`, or `settings` before constructing the typed tool definition.
- Add parameterized coverage for all three collisions.

#### Validation

- `uv run --no-sync pytest tests/python/test_sdk_contract.py -k typed_tool_definition`
- `just test-python`
- `uvx ruff format --check python/src/nemo_fabric/models.py tests/python/test_sdk_contract.py`
- `git diff --check`

Start with `ToolsConfig.add_definition()` in `python/src/nemo_fabric/models.py`; the regression test is directly below the existing typed-tool-definition coverage.

- Relates to: none

### DIFF
--- a/python/src/nemo_fabric/models.py
+++ b/python/src/nemo_fabric/models.py
@@ -982,11 +982,18 @@ class ToolsConfig(FabricBaseModel):
 
         if not name.strip():
             raise ValueError("tool definition names must not be empty")
+        extras = dict(extra_fields or {})
+        overlap = {"kind", "ref", "settings"}.intersection(extras)
+        if overlap:
+            raise ValueError(
+                "extra_fields duplicates known fields: "
+                + ", ".join(sorted(overlap))
+            )
         value = {
             "kind": kind,
             "ref": ref,
             "settings": dict(settings or {}),
-            **dict(extra_fields or {}),
+            **extras,
         }
         self.definitions[name] = ToolDefinitionConfig.model_validate(value)
         return self
--- a/tests/python/test_sdk_contract.py
+++ b/tests/python/test_sdk_contract.py
@@ -763,6 +763,19 @@ def test_typed_tool_definition_omits_empty_settings():
     }
 
 
+@pytest.mark.parametrize("field", ["kind", "ref", "settings"])
+def test_typed_tool_definition_rejects_known_extra_field_collisions(field: str):
+    tools = ToolsConfig()
+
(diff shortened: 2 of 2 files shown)


AnuradhaKaruppiah:
> Restore the NAT adapter's local model-role and MCP-server-name guards, plus registered workflow compatibility.

Hello! Can you explain the purpose of this change?
It seems to be in direct conflict with the new adapter approach -
1. NAT adapter shouldn't be doing validation (The invariants should be enforced once by the shared typed contract). That undermines the slim-adapter goal.
2. Why was "nat_workflow" added back? Directionally, we want to avoid NAT-specific custom-agent semantics. The workflow_settings.kind has to be a (a) well known type owned by the Fabric core (such as react, rewoo, tool-calling) or (b) a python-entrypoint or (c) python module. We don't want the adapter using a top level "kind=nat_workflow" escape hatch to interpret the config thereby bypassing the Fabric core's planning and validation.

external/nat/fabric-adapter.json
```diff
@@ -17,19 +17,40 @@
...
+            "additionalProperties": false
           },
-          "ref": {
-            "const": "fabric.agent.react",
-            "description": "Build Fabric's ReAct agent intent with the NAT react_agent factory."
+          {
+            "type": "object",
+            "properties": {
+              "kind": {
+                "const": "nat_workflow",
```
AnuradhaKaruppiah: we want to avoid having a toplevel escape hatch like "nat_workflow"

zhongxuanwang-nv:
@AnuradhaKaruppiah Thanks Anuradha for catching this!! This is definitely a huge regression and I definitely should not have missed it during my review — everywhere's saying the `kind` should only accept the three things you said, but I think the backward compat thing came up as one of the review issues and the agent took it; this is some area i should pay deep attention to in the future!!

AjayThorve:
I recommend narrowing this PR to only the `ToolsConfig.add_definition()` collision guard. The release-affecting result-schema issue has already landed in #224, and the other three changes are not required.

AjayThorve (approved):
Lgtm

PR #190, opened 2026-08-08
feat: add dependency-free adapter contract dataclasses

AnuradhaKaruppiah:
#### Overview

This change makes standard-library dataclasses the canonical Python representation of the southbound adapter contract. It removes Pydantic as an unconditional dependency of `nemo-fabric-adapter-contract` while retaining optional Pydantic interoperability through the `pydantic` extra.

The wire contract and published JSON Schemas do not change, so the adapter contract version remains unchanged. The dependency choice intentionally prefers the standard library for the base contract; a second Pydantic model hierarchy was rejected because duplicate field definitions could drift.

Breaking Python API changes:

- `model_validate(...)` becomes `from_mapping(...)`.
- `model_dump(mode="json")` becomes `to_mapping()`.
- Contract JSON Schema consumers use the published schemas or the optional Pydantic support module.
- The common lifecycle host replaces `config_model` with the framework-neutral `config_loader` callable.

#### Details

- Add a dependency-free codec with closed-object decoding, recursive typed construction, JSON-safe serialization, defensive extension copying, and field-path validation errors.
- Convert `AgentConfig`, `AgentRunRequest`, `AgentRunResult`, `RuntimeContext`, and their nested blocks to keyword-only, slotted dataclasses.
- Add optional Pydantic `TypeAdapter` and typed-extension helpers without duplicating contract classes.
- Make the common lifecycle host accept a generic configuration loader.
- Move the NAT reference adapter to `AgentConfig.from_mapping`.
- Follow-up: as first-party adapters migrate to `AgentConfig`, add a direct `nemo-fabric-adapter-contract == 0.2.0` dependency to each adapter package. The base contract package has no runtime dependencies; Pydantic is installed only through the optional `[pydantic]` extra.
- Add dependency-boundary, schema-parity, packaging, lifecycle, and NAT coverage.
- Align `ArtifactRef` with the published runtime-context schema by including its existing `metadata` field.
- Update package guidance, adapter contract documentation, and the public adapter-authoring skill.

#### Validation

- `just --set no_uv true test-python` — 713 passed, 15 skipped.
- Focused NAT tests — 49 passed, 1 skipped.
- Focused lifecycle tests — 11 passed.
- Built the adapter-contract wheel and installed it with `--no-deps` in a clean Python 3.11 environment; `AgentConfig` round-tripped without importing Pydantic.
- Ruff 0.15.21 formatting and checks passed for changed Python files.
- Root and adapter-contract `uv lock --check` passed.
- `git diff --check` passed.
- Python license diff completed. The full documentation build was not run; documentation tests in the Python suite passed.

Start with `adapter-contract/src/nemo_fabric_adapter_contract/models.py` and `codec.py`. The central design decision is that the dataclasses are the only contract hierarchy, while `pydantic_support.py` adapts those same types when a consumer opts into Pydantic.

Then review `adapters/common/src/nemo_fabric_adapters/common/lifecycle.py` and the NAT adapter migration to confirm the framework-neutral `config_loader` boundary.

- Relates to #186

## New Features
- Added dependency-free configuration dataclasses with strict validation and mapping serialization.
- Added detailed validation errors with nested field paths.
- Added artifact metadata support.
- Added optional Pydantic interoperability for typed extensions and schema generation.

## Breaking Changes
- Lifecycle configuration now uses `config_loader` instead of `config_model`.

## Documentation
- Updated adapter setup, configuration, and lifecycle guidance for the new APIs.
- Clarified that Pydantic is optional rather than required.

### DIFF
--- a/adapter-contract/src/nemo_fabric_adapter_contract/codec.py
+++ b/adapter-contract/src/nemo_fabric_adapter_contract/codec.py
@@ -141,6 +141,19 @@ def validate_dataclass(instance: Any) -> None:
         object.__setattr__(instance, item.name, decoded)
 
 
+def decode_field(instance: Any, name: str, value: Any) -> Any:
+    """Validate and normalize one field assignment."""
+
+    item = next(item for item in fields(instance) if item.name == name)
+    annotation = _resolved_type_hints(type(instance))[name]
+    return _decode_value(
+        annotation,
+        value,
+        path=(name,),
+        field=item,
+    )
+
+
 def encode_dataclass(instance: Any) -> dict[str, Any]:
     """Return a detached JSON-compatible mapping for a contract dataclass."""
 
--- a/adapter-contract/src/nemo_fabric_adapter_contract/codec.py
+++ b/adapter-contract/src/nemo_fabric_adapter_contract/codec.py
@@ -1,0 +1,273 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Dependency-free encoding and validation for adapter contract dataclasses."""
+
+from __future__ import annotations
+
+import math
+import types
+from collections.abc import Mapping
+from dataclasses import MISSING
+from dataclasses import Field
+from dataclasses import fields
+from enum import Enum
+from pathlib import Path
+from typing import Any
+from typing import Literal
+from typing import TypeVar
+from typing import Union
+from typing import get_args
+from typing import get_origin
+from typing import get_type_hints
+
+
+# JSON's recursive type cannot be expressed as a named standard-library alias
+# on every supported Python version. Contract fields marked as JSON are checked
+# recursively by ``json_value`` at runtime.
+JsonValue = Any
+
+_T = TypeVar("_T")
+
+
+class ContractValidationError(ValueError):
+    """A southbound contract value failed dependency-free validation."""
+
+    def __init__(self, message: str, *, path: tuple[str, ...] = ()) -> None:
+        self.message = message
+        self.path = path
+        location = ".".join(path)
+        super().__init__(f"{location}: {message}" if location else message)
+
+    def prepend(self, path: tuple[str, ...]) -> "ContractValidationError":
+        """Return the same validation failure beneath an outer field path."""
+
+        return ContractValidationError(self.message, path=(*path, *self.path))
+
+
+def json_value(value: Any, *, path: tuple[str, ...] = ()) -> JsonValue:
+    """Validate, copy, and return one JSON-compatible value."""
+
+    if value is None or isinstance(value, (bool, str)):
+        return value
+    if isinstance(value, int):
+        return value
+    if isinstance(value, float):
+        if not math.isfinite(value):
+            raise ContractValidationError("must be a finite JSON number", path=path)
+        return value
+    if isinstance(value, list):
+        return [
+            json_value(item, path=(*path, str(index)))
+            for index, item in enumerate(value)
+        ]
+    if isinstance(value, Mapping):
+        result: dict[str, JsonValue] = {}
+        for key, item in value.items():
+            if not isinstance(key, str):
+                raise ContractValidationError(
+                    "JSON object keys must be strings", path=path
+                )
+            result[key] = json_value(item, path=(*path, key))
+        return result
+    raise ContractValidationError("must be a valid JSON value", path=path)
+
+
+def json_mapping(
+    value: Mapping[str, Any], *, path: tuple[str, ...] = ()
+) -> dict[str, JsonValue]:
+    """Validate and detach one JSON object mapping."""
+
+    result = json_value(value, path=path)
+    if not isinstance(result, dict):  # pragma: no cover - kept true by the annotation
+        raise ContractValidationError("must be a JSON object", path=path)
+    return result
+
+
+def decode_dataclass(model: type[_T], value: Any, *, path: tuple[str, ...] = ()) -> _T:
+    """Decode one closed mapping into a contract dataclass."""
+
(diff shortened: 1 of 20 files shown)


adapter-contract/src/nemo_fabric_adapter_contract/codec.py
```diff
@@ -0,0 +1,281 @@
...
+    if annotation is int:
+        if type(value) is not int:
+            raise ContractValidationError("must be an integer", path=path)
+        return value
+    if annotation is float:
+        if isinstance(value, bool) or not isinstance(value, (int, float)):
+            raise ContractValidationError("must be a number", path=path)
+        result = float(value)
+        if not math.isfinite(result):
+            raise ContractValidationError("must be a finite number", path=path)
```
zhongxuanwang-nv: Very nitty thing, should we also catch `OverflowError`, for `float(value)`, before checking `math.isfinite(result)`?

adapter-contract/src/nemo_fabric_adapter_contract/pydantic_support.py
```diff
@@ -0,0 +1,42 @@
...
+from nemo_fabric_adapter_contract.models import AgentContractBlock
+
+
+_T = TypeVar("_T")
+
+
+def type_adapter(model: type[_T]) -> TypeAdapter[_T]:
+    """Return a Pydantic adapter for one canonical contract dataclass."""
+
+    return TypeAdapter(model)
```
zhongxuanwang-nv, line 24: `type_adapter(AgentModelConfig).json_schema()` now emits `provider` as a bare string and `max_turns` with no bound. Should we extend the tests (`test_agent_config.py:90-99`, `test_agent_execution.py:108-121`) to compare constraints — minLength/pattern/minimum/maximum — against the Rust schemas, not just property names?

adapter-contract/src/nemo_fabric_adapter_contract/models.py
```diff
@@ -1,311 +1,302 @@
...
+@dataclass(slots=True, kw_only=True)
+class ContractModel:
+    """Base for adapter-facing contract dataclasses."""
+
+    # Pydantic reads this metadata only when its optional TypeAdapter is used.
+    # It does not require importing Pydantic in the base contract package.
+    __pydantic_config__: ClassVar[dict[str, Any]] = {
+        "extra": "forbid",
+        "allow_inf_nan": False,
+    }
```
zhongxuanwang-nv, line 75: Nitty thing too, the old `ContractModel` carried `ConfigDict(extra="forbid", validate_assignment=True, allow_inf_nan=False)`, but here it only validates in `__post_init__` so post-construction assignment is unchecked, e.g.

`u = AgentUsage(input_tokens=1); u.input_tokens = -5; u.to_mapping()` returns `{'input_tokens': -5}` (the schema requires >= 0), and flipping `r.status = AgentRunStatus.FAILED` serializes a failed result with no `error`, violating the class's own invariant.

Therefore, should we restore the always-valid invariant, by doing things like `frozen=True` on the contract dataclasses with `dataclasses.replace(...)` for updates (requires updating the tests that mutate).

Or, we can have a `__setattr__` guard on ContractModel that, after init, re-runs the field decode and `self._validate()` on assignment?

zhongxuanwang-nv (approved):
Thanks Anuradha!!

adapter-contract/src/nemo_fabric_adapter_contract/codec.py
AnuradhaKaruppiah: makes sense

adapter-contract/src/nemo_fabric_adapter_contract/pydantic_support.py
AnuradhaKaruppiah: I am not sure if we need to overrotate on testing the pydantic schemas :)

adapter-contract/src/nemo_fabric_adapter_contract/models.py
AnuradhaKaruppiah: nice! so invalid values can be serialized :) let me see how to fix this...

AnuradhaKaruppiah: I am going to allow changes (no freezing) and re-validate the entire object as a part of to_mapping

adapter-contract/src/nemo_fabric_adapter_contract/codec.py
AnuradhaKaruppiah: Addressed in 94443c1. The float decoder now converts OverflowError into ContractValidationError with the field path, with regression coverage for an oversized integer temperature.

adapter-contract/src/nemo_fabric_adapter_contract/models.py
AnuradhaKaruppiah: Addressed in 94443c1 and optimized in 0906c71. ContractModel now decodes and validates only the assigned field, then runs the model cross-field validation. A failed assignment is rolled back, preserving valid mutation and set_extensions(...) ergonomics without revalidating unrelated fields or nested collections. Regression coverage includes both reported examples.

adapter-contract/src/nemo_fabric_adapter_contract/models.py
```diff
@@ -1,311 +1,317 @@
...
+    def __post_init__(self) -> None:
+        validate_dataclass(self)
+        self._validate()
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        try:
+            previous = getattr(self, name)
+        except AttributeError:
+            object.__setattr__(self, name, value)
+            return
```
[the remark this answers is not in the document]
AnuradhaKaruppiah: Addressed in 98fbd7c. ContractModel now rejects names outside the declared dataclass field set before the initialization path, with regression coverage for post-construction assignment of an unknown AgentUsage attribute.

PR #191, opened 2026-08-10
fix: keep a completed Deep Agents turn completed when telemetry fails

SandyChapman:
#### Overview

`DeepAgentsRuntime.invoke()` wrapped the Relay plugin/scope context managers and the agent call in a single `try/except Exception`. A context manager's `__exit__` runs inside that `try`, so an exception raised purely while *closing* the telemetry scope was assigned to the invocation `error`, and `normalize_output` then reported `completed=false` / `failed=true` for a turn the agent had already finished — final response, tool calls, and workspace changes all intact.

This is reachable today. NeMo Relay keeps one process-global LIFO scope stack in a `ContextVar`; LangGraph schedules child tasks with `copy_context()`, which preserves that *same mutable* stack rather than cloning it, so concurrent chain callbacks can close out of LIFO order and Relay's scope validator rejects the pop with:

```
RuntimeError: invalid argument: scope handle is not at the top of the stack
```

The stranded child scope then makes the outer `deepagents-request` scope raise on exit. Downstream this scored 31 otherwise-successful NeMo Evaluator agent-eval trials as adapter invocation failures (NVBug 6562846), including at evaluator parallelism 1. Deep Agents is where this is reachable because it is the only adapter that installs a LangChain callback handler, and its nested per-chain scopes are what LangGraph's concurrent tasks close out of order.

This PR separates the two failure domains structurally rather than papering over the Relay bug:

- **`TurnOutcome`** carries `error` (the agent failed, no functional result) and `telemetry_error` (recording the turn failed) as independent fields.
- **`_invoke_agent`** owns the invocation domain. **`_invoke_with_telemetry`** owns the telemetry one and guards nothing but the context managers, so an exception it catches can only be a telemetry fault.
- Whether the agent got to run is answered by whether an outcome exists, not by a side flag. No outcome means telemetry failed on the way *in* and there is nothing to preserve, so that stays an invocation failure — the split cannot widen into general failure suppression.
- **Artifact collection is covered too.** It walks the filesystem *after* the turn is over, so letting it raise would discard a completed invocation — the same failure mode this change exists to stop. `_telemetry_output` returns that fault instead of raising, and faults from multiple stages are joined rather than the first one winning.
- **`telemetry.degraded: true`** accompanies `telemetry.error`, so a consumer has a machine-readable signal instead of a message to parse. Artifacts are still referenced when degraded — a partial trajectory is usually worth reading — but the flag marks them as possibly truncated. Both keys are absent on a clean run, and the block is emitted for a fault even without a runtime block so the diagnostic cannot be silently dropped.

The Relay-side root cause (shared scope stack across concurrent LangGraph tasks; a failed `scope.pop` stranding an untracked scope) is being reported separately against NeMo-Relay. This change is worth making on its own regardless: an observability fault should never rewrite the functional outcome of a completed invocation.

`adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py` — `TurnOutcome` and the `_invoke_with_telemetry` / `_invoke_agent` split. The design decision to check is the `outcome is None` test in `_invoke_with_telemetry`: it is what distinguishes "telemetry broke on the way out, after a completed turn" (recoverable) from "telemetry broke on the way in, so the agent never ran" (a real failure).

The contract question worth a maintainer's opinion is the artifact policy: on a degraded run this **references artifacts anyway** and flags them, rather than withholding them. Withholding is the other defensible choice; happy to switch if you'd rather a degraded run expose nothing.

Then `tests/adapters/test_deepagents.py`, which pins each branch: scope-teardown fault keeps the turn completed; plugin-flush fault likewise; artifact-collection fault likewise; two faults in one turn are both reported; a setup fault stays an invocation failure; a genuine agent failure still fails with a clean telemetry block; and agent failure plus teardown fault keep their own domains.

Review follow-ups from this round: the completed-result claim in the README was overstated (a telemetry fault after an agent *failure* correctly leaves the turn failed) and is now qualified; the setup-failure test asserts `invoke_compiled_agent` was never awaited, so a regression that consumes an agent request before reporting the fault is caught.

Not applied: redacting the exception text in `telemetry.error`. The concern is real, but this adapter already returns raw `str(exc)` in the invocation `error` and has since before this PR, and that path is the likelier credential carrier (a provider client error can echo a `base_url`). Redacting only the new field would be inconsistent and would remove the diagnostic the field exists for. If the exception text should be sanitized, it belongs in one place across the normalized result — happy to open that separately.

Testing:

- `just test-python` — 680 passed, 53 skipped.
- `tests/adapters/test_deepagents.py` — 49 passed (42 before, 7 new).
- Mutation checks, each caught by a test: forcing the outer handler to always keep the outcome (drops the setup-failure branch); letting artifact collection raise; reporting only the first of two faults.
- `ruff check` / `ruff format --check` (v0.15.21, matching `.pre-commit-config.yaml`) and the copyright-header hook pass on all changed files.

Not addressed here, flagged rather than fixed: `_invoke_agent` collapses an exception to `f"{type}: {exc}"` and discards the traceback, which is pre-existing behavior and this module has no logger to route it to. Separately, `tests/adapters/test_claude_adapter.py::test_build_options_maps_normalized_capabilities_and_claude_settings` fails whenever `ANTHROPIC_BASE_URL` is set in the environment — unrelated to this PR, but it bites anyone running the suite from a shell that exports it, and a `monkeypatch.delenv` would fix it.

No breaking changes: the normalized result gains optional `telemetry.degraded` / `telemetry.error` keys and is otherwise unchanged, and `normalize_output`'s new `telemetry_error` parameter is keyword-only with a default.

- Relates to NVBug 6562846

## Summary by CodeRabbit

- **Bug Fixes**
  - Completed agent responses remain successful when telemetry teardown or artifact collection fails.
  - Telemetry issues are reported separately with degraded status and error details.
  - Failures during telemetry setup or before agent execution continue to return invocation errors.
  - Agent execution failures remain invocation errors, while later telemetry failures are reported independently.
  - Partial or unavailable telemetry artifacts no longer replace completed functional responses.

- **Documentation**
  - Clarified telemetry failure handling across setup, execution, teardown, and artifact collection.

### DIFF
--- a/adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
+++ b/adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
@@ -50,6 +50,19 @@ FABRIC_OWNED_AGENT_KEYS = frozenset(
 # through harness.settings.deepagents. Executable objects (AgentMiddleware, BaseTool,
 # Python callables) cannot cross the SDK->JSON->payload boundary and are excluded.
 DEEPAGENTS_PASSTHROUGH_KEYS = frozenset({"subagents", "interrupt_on"})
+# Appended to the fault that poisoned Relay's scope stack, and then reported on every
+# later turn of the same runtime so none of them can look telemetry-clean. Deliberately
+# does not claim later turns are untraced: the Relay middleware is attached to the
+# compiled agent at start and keeps emitting, so what is actually lost is trustworthy
+# nesting, not all telemetry.
+_QUARANTINE_NOTE = (
+    "telemetry unreliable for the rest of this runtime: an earlier turn left the Relay "
+    "scope stack dirty, so this turn is not wrapped in a request scope and any events "
+    "the agent middleware still emits are nested under a stale scope"
+)
+# Sentinel for "this handle carries no identity", kept distinct from a real ``None``
+# attribute value so an unreadable handle can never compare equal to another one.
+_UNREADABLE = object()
 
 
 class AdapterConfigError(RuntimeError):
@@ -471,6 +484,8 @@ class DeepAgentsRuntime:
         self._relay_scope_type: Any = None
         self._relay_plugin_config: dict[str, Any] | None = None
         self._callback_handler_type: Any = None
+        self._telemetry_quarantine: str | None = None
+        self._telemetry_quarantine_cause: str | None = None
 
     async def start(self, payload: dict[str, Any]) -> None:
         if self._started:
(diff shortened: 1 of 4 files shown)


adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
```diff
@@ -588,20 +595,26 @@ async def invoke(self, invocation: dict[str, Any]) -> dict[str, Any]:
...
+                    turn_messages,
+                    error,
+                ) = await self._invoke_agent(user_message)
+                invoked = True
+        except Exception as exc:  # telemetry lifecycle fault
+            telemetry_error = f"{type(exc).__name__}: {exc}"
+            if not invoked:
+                # Telemetry failed before it ever handed control to the agent, so there
+                # is no functional outcome to preserve; this stays an invocation failure.
+                error = telemetry_error
```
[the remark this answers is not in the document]
SandyChapman: Not applying: `error` already returns raw `str(exc)` on main (line 604), so redacting only the field this PR adds leaves the bigger surface open while looking fixed. If exception text should be sanitized it should cover both in one place — happy to open that separately.

SandyChapman:
The 41.38% is measured over the whole file rather than the diff; 5 of the 6 functions this PR adds have docstrings. Raising the full module to 80% would be a large unrelated diff, but happy to do it as a follow-up.

adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
```diff
@@ -570,73 +570,119 @@ async def invoke(self, invocation: dict[str, Any]) -> dict[str, Any]:
...
 
-        result_state: Any = None
-        events: list[dict[str, Any]] = []
-        turn_messages: list[dict[str, Any]] = []
-        error: str | None = None
         resumed = self._completed_invocations > 0
+        if self._observability is None:
+            outcome = await self._invoke_agent(user_message)
+        else:
+            outcome = await self._invoke_with_telemetry(user_message, request_id)
```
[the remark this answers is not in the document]
SandyChapman: Pre-existing — `main` routes on the same `self._observability is not None` condition and this PR does not touch it, so native already goes through the Relay path. Real bug, but rewiring telemetry routing is out of scope for this fix.

adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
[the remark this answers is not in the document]
SandyChapman: No follow-up needed, thanks — I am an outside contributor here, so I will leave that call to the maintainers.

adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
```diff
@@ -570,73 +570,119 @@ async def invoke(self, invocation: dict[str, Any]) -> dict[str, Any]:
...
+                    scope_error = _error_text(exc)
+        except Exception as exc:  # telemetry lifecycle fault
+            telemetry_error = _join_faults(scope_error, _error_text(exc))
+        else:
+            telemetry_error = scope_error
+
+        if outcome is None:
+            # No outcome means the agent never ran, so there is nothing to preserve.
+            return TurnOutcome(error=telemetry_error, telemetry_error=telemetry_error)
+        return outcome._replace(telemetry_error=telemetry_error)
```
AjayThorve, line 669: [P1] Keep a telemetry teardown fault from producing a false-clean next turn

Catching this failed outer pop preserves the completed turn, but it does not restore Relay's ContextVar scope stack. I reproduced two invokes on the same DeepAgentsRuntime: turn 1 overlaps A/B and returns completed=true/degraded=true while leaving A as the current scope; turn 2 is properly nested yet returns completed=true with no telemetry.degraded/error while scope.get_handle() is still A from turn 1. Relay plugin cleanup does not reset this stack, and the adapter host reuses one event-loop context across invoke*.

Please establish a post-fault invariant before returning: restore a supported clean Relay context, or quarantine/recreate telemetry and keep degradation sticky until recovery. This does not require implementing the Relay concurrency fix here, but a known-corrupt runtime must not report a later turn as telemetry-clean. Please cover the behavior with two ordered invokes.

tests/adapters/test_deepagents.py
```diff
@@ -386,6 +386,285 @@ async def test_relay_telemetry_wraps_agent_and_reports_artifacts(
...
+            "relay_config": {},
+            "native_config": None,
+            "adapter_outputs": [],
+        }
+        return payload
+
+    return build
+
+
+async def test_relay_scope_teardown_failure_keeps_the_invocation_completed(
```
AjayThorve, line 453: [P2] Exercise the real callback overlap and persistent state

This monkeypatch proves result normalization when the outer scope raises, but it cannot detect the reported state leak: callback A is removed from _scope_handles before its pop fails, A remains on the shared stack, and the next invocation reuses that stack.

Please add an integration regression using the actual Relay callback/scope sequence—A start, B start, A end, B end—followed by another invoke. Assert that callback tracking is empty, no stale scope remains current, and a poisoned runtime cannot be reported telemetry-clean.

AjayThorve (changes requested):
The agent-outcome/telemetry-outcome split is the right containment boundary, and the real-agent runs confirm that response and workspace effects survive the teardown fault. I am requesting changes only for cross-invocation containment: the first degraded turn currently leaves Relay state corrupted while a later turn can be reported clean.

The concurrency/root fix likely belongs in NeMo Relay, not in this adapter. Current Relay 0.7.x has task-context forking primitives from [NVIDIA/NeMo-Relay#589](https://github.com/NVIDIA/NeMo-Relay/pull/589), but its LangChain callback does not apply them at overlapping sibling-run boundaries. This Fabric PR should establish a safe post-fault invariant and link/track the Relay-side fix rather than reimplementing Relay internals here.

The full NVBug acceptance still requires the Relay fix, complete ATOF/ATIF without scope-pop errors, and the Platform test matrices. Once the persistent-runtime invariant below is addressed, I would be comfortable merging this as containment while the Relay fix proceeds.

adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
SandyChapman: Done — a fault that leaves a scope current now quarantines the runtime: later turns open no request scope, reference no artifacts of their own, and always report degraded. It survives `stop()`/`start()` since the stack is process-scoped, and it is covered with two ordered invokes.

tests/adapters/test_deepagents.py
SandyChapman: Added `tests/integrations/test_relay_scope_leak.py` — the real A/B/A-end/B-end callback overlap, plus two ordered turns through the adapter's telemetry path proving a poisoned runtime cannot report clean. One thing I deliberately did not do: assert no stale scope remains current. Containment cannot clean the stack, so that test asserts the leak instead; the restore belongs in the Relay fix.

adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
```diff
@@ -570,73 +585,138 @@ async def invoke(self, invocation: dict[str, Any]) -> dict[str, Any]:
...
         }
-        relay_artifacts = (
-            common_utils.collect_relay_artifacts(self._observability.plugin_config)
-            if self._observability.collect_artifacts
-            else None
-        )
-        return telemetry_runtime, relay_artifacts
+        if not self._observability.collect_artifacts:
+            return telemetry_runtime, None, None
+        if self._telemetry_quarantine is not None:
```
AjayThorve, line 711: [P2] Keep the faulting turn's artifact policy consistent

When this turn detects a dirty scope, _invoke_with_telemetry() sets _telemetry_quarantine before invoke() calls _telemetry_output(). This branch therefore suppresses relay_artifacts for the faulting turn too. I reproduced faulting_turn_artifacts=None, while the README and PR body say a scope/flush-degraded turn still references its partial artifacts; only later turns that inherit quarantine should omit their own artifacts.

Non-blocking for this containment fix, but please either pass the pre-invocation inherited_quarantine state into artifact collection so the poisoning turn can publish its partial artifacts, or update the documented contract and regression test if withholding is intended.

AjayThorve (approved):
Validated the current head: the persistent-runtime containment now keeps poisoned runtimes degraded across ordered invokes and stop/start, and the real Relay callback regression covers the A/B/A-end/B-end leak. Focused validation passed 60 tests; just test-python passed 730 tests with 16 skipped; hosted checks are green.

Approving this as the Fabric containment fix. The root scope-stack repair remains owned by NeMo Relay. I left one non-blocking P2 inline for the faulting-turn artifact contract.

PR #194, opened 2026-08-10
feat: Support MCP Authentication

dagardner-nv:
#### Overview

* Adds support for MCP Oauth2 as well as Oauth 2 service accounts
* Adds `custom_headers` field to `McpServerConfig` allowing for static API Key based auth
* Adds an `mcp-oauth` extra to `nemo-fabric-adapters-common`
* Common MCP Auth methods in `adapters/common/src/nemo_fabric_adapters/common/mcp_auth.py`
* MCP transport is now an enum (stdio, sse, streamable-http)

Each harness has it's own support (or lack of) MCP Auth from the SDK, however most assume the user will run `<harness> mcp login <mcp name>` and authenticate an MCP server globally. 

* Claude Code
  + Claude doesn't offer a way to authenticate an MCP server with the SDK, and running `claude mcp login` in a subprocess requires an interactive terminal.
  + Oauth is handled by the Fabric adapter and then hands authentication tokens to Claude.
  + Tokens are stored as specially named environment variables in the form of `NEMO_FABRIC_CLAUDE_MCP_<sha256 hash>`
  + An `mcp.json` file is written to disk referencing the environment variables ensuring credentials are never written to disk, ex:
  ```json
  "Authorization": "Bearer ${NEMO_FABRIC_CLAUDE_MCP_5EF663A783BD47F}"
  ```

* Codex
  + From `CodexRuntime.invoke` Fabric queries Codex for the auth status of each MCP server
  + For those that are not authenticated Fabric makes a call to the Codex JSON-RPC `mcpServer/oauth/login` endpoint.

* Deepagents
  + Fabric creates an Oauth provider which it then passes to Deepagents
  + Deepagents invokes the Oauth provider as needed

* Hermes Agent
  + Un-athenticated MCP servers are ignored by default when run in a non-interactive terminal. 
  + from `HermesRuntime.invoke` a call is made to `tools.mcp_tool.get_mcp_status` if there are any MCP servers marked as disconnected, a call is made to  `tools.mcp_oauth.force_interactive_oauth` 
  + Note: the `tools` namespace is provided by the `hermes-agent` Python package.

* `crates/fabric-core/src/config.rs`

- Closes FABRIC-170

## Summary by CodeRabbit

* **New Features**
  * Added OAuth 2.0 and service-account authentication for MCP servers.
  * Added custom HTTP headers, browser sign-in, secure credential handling, token caching, and automatic refresh.
  * Added typed MCP transport options: stdio, SSE, and streamable HTTP.
  * Extended Python and Rust SDK configuration and schemas with authentication settings.

* **Documentation**
  * Updated SDK and API references with MCP authentication and transport options.

* **Tests**
  * Added coverage for authentication flows, validation, token refresh, headers, and adapter integration.


adapters/claude/src/nemo_fabric_adapters/claude/adapter.py
```diff
@@ -324,10 +325,178 @@ def _mcp_servers(payload: dict[str, Any]) -> dict[str, Any]:
...
                 f"unsupported MCP transport: {transport}",
             )
+        if headers := server.get("custom_headers"):
+            try:
+                result[name]["headers"] = mcp_auth.normalize_custom_headers(
+                    name, headers
+                )
+            except mcp_auth.McpAuthConfigError as error:
+                raise AdapterConfigError(
+                    "claude_invalid_configuration", str(error)
```
zhongxuanwang-nv, line 335: Nitty thing here — `custom_headers` values here would be written to the `mcp.json` and would bypass the `NEMO_FABRIC_CLAUDE_MCP_<sha256>` env projection that `_stage_mcp_config` applies to env values and OAuth tokens; would this be an expected behavior?

zhongxuanwang-nv: Similar thing for the hermes adapter

adapters/claude/src/nemo_fabric_adapters/claude/adapter.py
```diff
@@ -1037,19 +1249,72 @@ async def invoke(self, invocation: dict[str, Any]) -> dict[str, Any]:
...
 
+    async def _authenticate_mcp_servers(
+        self,
+        payload: dict[str, Any],
+        client: ClaudeSDKClient,
+        invocation_deadline: float,
+    ) -> None:
+        for name in _authenticated_mcp_servers(payload):
+            if self._mcp_authentication_checked.get(name, False):
+                continue
```
zhongxuanwang-nv, line 1260: The OAuth provider and the refresh token it obtained are discarded after `_prefetch_mcp_oauth_tokens` projects a one-shot static access token into `mcp.json` at start, and `_authenticate_mcp_servers` skips servers that are already marked in `_mcp_authentication_checked`, so an expired token may never detected or refreshed for the life of the runtime?

adapters/claude/src/nemo_fabric_adapters/claude/adapter.py
```diff
@@ -324,10 +325,178 @@ def _mcp_servers(payload: dict[str, Any]) -> dict[str, Any]:
...
+    except mcp_auth.McpAuthConfigError as error:
+        raise ClaudeAdapterError(
+            "claude_mcp_authentication_failed",
+            f"MCP server {server_name!r} OAuth configuration is invalid: {error}",
+            metadata={"server": server_name},
+        ) from error
+
+    try:
+        async with asyncio.timeout(timeout):
+            async with httpx.AsyncClient(auth=provider) as http_client:
```
zhongxuanwang-nv, line 424: Should we also include `custom_headers` here, so servers requiring both a static header and OAuth never issue the 401 challenge?

adapters/deepagents/pyproject.toml
```diff
@@ -25,7 +25,7 @@ license-files = ["LICENSE"]
 readme = "pypi.md"
 requires-python = ">=3.11"
 dependencies = [
-  "nemo-fabric-adapters-common == 0.2.0",
```
zhongxuanwang-nv, line 28: Should we also declare the same thing for Claude adapter too, because Claude adapter also uses `mcp_auth.create_mcp_oauth_provider` which is in this `mcp-oauth` extra?

crates/fabric-core/src/config.rs
```diff
@@ -707,19 +707,128 @@ pub struct McpConfig {
...
+            skip_serializing_if = "is_default_mcp_token_cache_buffer_seconds"
+        )]
+        token_cache_buffer_seconds: u64,
+    },
+}
+
 /// MCP server configuration.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
 pub struct McpServerConfig {
     /// MCP transport.
```
zhongxuanwang-nv, line 819: Nitty thing, but should we market this PR as a breaking change too?

adapters/hermes/src/nemo_fabric_adapters/hermes/adapter.py
```diff
@@ -222,9 +227,13 @@ async def invoke(self, invocation: dict[str, Any]) -> dict[str, Any]:
             user_message = json.dumps(user_message, sort_keys=True)
         instructions = agent_config.instructions
         system_prompt = (
-            instructions.system.content if instructions and instructions.system else None
+            instructions.system.content
+            if instructions and instructions.system
+            else None
         )
 
+        await self._authenticate_mcp_servers()
```
zhongxuanwang-nv, line 235: This is a review left by agent that I didn't verify, because I feel like this is exactly something that an agent would be really good at to verify if the claim is true 😂

"Hermes _authenticate_mcp_servers raises lifecycle.LifecycleError out of invoke(), which the shared lifecycle host wraps as _AdapterCallError and records as state.failed=True (lifecycle.py:363-365), permanently rejecting all later invokes with lifecycle_runtime_failed (lifecycle.py:258-260) — so the retry intent encoded by leaving self._mcp_authentication_checked False on failure (asserted by test_runtime_reports_failed_oauth_mcp_authentication) is unreachable, unlike Claude and Codex which return structured failure outputs for the same condition and stay retryable."

zhongxuanwang-nv (commented):
Thanks David!! Used powerful AI to help me review, and this is the first round, as i just realized this PR is in draft now!

adapters/deepagents/pyproject.toml
zhongxuanwang-nv: oops it did not catch the next line loll, but it is `  "nemo-fabric-adapters-common[mcp-oauth] == 0.2.0",`

adapters/hermes/src/nemo_fabric_adapters/hermes/configuration.py
```diff
@@ -138,13 +140,17 @@ def write_hermes_config(
...
     target = os.path.expandvars(server.url).strip()
-    if not target:
-        raise ValueError("MCP server mapping requires a URL")
 
     if transport == "stdio":
+        if server.authentication:
+            raise ValueError("MCP authentication is not supported for stdio transport")
+        if server.custom_headers:
+            raise ValueError("MCP custom_headers are not supported for stdio transport")
         return common_utils.without_none(
```
zhongxuanwang-nv, line 148: is the removal intentional?

zhongxuanwang-nv (commented):
oops this was supposed to get in for the last round

adapters/claude/src/nemo_fabric_adapters/claude/adapter.py
dagardner-nv: In short no, this will need to be a documentation item for Claude Code, as `custom_headers` is serialized to disk doing:
```py
config.add_mcp_server(..., authentication=None, custom_headers={"Authorization": "Bearer XYZ"})
```

Will cause the token to be serialized to disk (it is created with permissions set to 600, but this isn't supported on Windows, and the file is remove on a graceful shutdown, but will remain on disk if the process exits unexpectedly like receiving a SIGKILL).

Instead the more secure approach would be:
```py
config.add_mcp_server(..., authentication=None, custom_headers={"Authorization": "Bearer ${MCP_ACCESS_TOKEN}"}, env={"MCP_ACCESS_TOKEN": "${MCP_ACCESS_TOKEN}"})
```

Which will avoid having the token serialized to disk.

PR #198, opened 2026-08-11
feat: move Codex adapter to v1alpha2

zhongxuanwang-nv:
#### Overview

Move the Codex adapter to the `fabric.adapter/v1alpha2` typed contract. Lifecycle startup validates `AgentConfig`; each helper receives only the typed configuration, `RuntimeContext`, and base directory it needs, while model, MCP, skills, instructions, and telemetry are read directly from the contract.

The runtime telemetry context now carries the already-resolved native telemetry config required by Codex. Relay keeps its prior precedence rules through the shared typed model-name override. No shared helper is removed: caller audit still finds Claude, Codex, Hermes, and Deep Agents users.

Validated with `pytest -q tests/adapters/test_codex_adapter.py` (47 passed), `just test-python` (705 passed, 56 skipped), and `cargo test -p nemo-fabric-core runtime_context_preserves_native_telemetry_config --lib` (1 passed).

Start with `adapters/codex/src/nemo_fabric_adapters/codex/adapter.py`, then `crates/fabric-core/src/runtime.rs` for the native telemetry handoff.

- Relates to [FABRIC-185](https://linear.app/nvidia/issue/FABRIC-185/move-codex-adapter-to-v1alpha2-contract-version)


adapters/common/src/nemo_fabric_adapters/common/utils.py
```diff
@@ -325,12 +327,12 @@ def load_relay_plugin_config(payload: dict[str, Any]) -> dict[str, Any]:
...
+    normalize_relay_output_dirs(plugin_config, payload, model_name=model_name)
     return plugin_config
 
 
 def normalize_relay_output_dirs(
-    plugin_config: dict[str, Any], payload: dict[str, Any]
+    plugin_config: dict[str, Any], payload: dict[str, Any], *, model_name: str | None = None
 ) -> None:
     base = Path(base_dir(payload)).resolve()
     runtime_id = runtime_context(payload)["runtime_id"]
```
AnuradhaKaruppiah, line 336: The caller has already decoded a validated `RuntimeContext`, but this helper still reparses `runtime_context` from the raw lifecycle payload. Since this migration is intended to move contract-owned data off defensive JSON parsing, please pass the validated `runtime_id` (and the other envelope values this helper needs) explicitly. A legacy fallback can remain for unmigrated adapters if necessary.

tests/adapters/test_codex_adapter.py
```diff
@@ -14,18 +14,45 @@
...
+        **payload,
+        "config": config,
+        "runtime_context": context,
+        "request": None,
+    }
+
+
+def runtime_input(payload):
+    context = dict(payload["runtime_context"])
+    telemetry_plan = payload.get("telemetry_plan")
```
AnuradhaKaruppiah: This test helper reconstructs the typed telemetry context from the legacy `telemetry_plan`, and several fixtures similarly populate both `capability_plan` and `config`. That lets the migration tests succeed while test code performs the legacy-to-typed translation. Please author these fixtures directly with `AgentConfig`/`RuntimeContext` wire data, and add a conflicting-legacy-values case proving that only the typed contract is authoritative.

adapters/codex/src/nemo_fabric_adapters/codex/adapter.py
```diff
@@ -1293,44 +1297,44 @@ async def invoke(self, invocation: dict[str, Any]) -> dict[str, Any]:
                 "codex_runtime_not_started",
                 "Codex runtime is not started",
             )
-        if runtime_id(invocation) != self._fabric_runtime_id:
+        config = self._config
+        context = self._context
```
AnuradhaKaruppiah, line 1314: The current invocation `RuntimeContext` is decoded below but is used only for the runtime-ID check; `_invoke_thread` and result normalization receive this cached start-time context. `RuntimeContext` is invocation-scoped (`invocation_id`, `request_id`, artifacts, and telemetry), so please pass the newly decoded context through the invocation path. A two-turn test with distinct context values would make the contract authority explicit.

tests/adapters/test_codex_adapter.py
zhongxuanwang-nv: oops, would remove these and de-dup!

AnuradhaKaruppiah (approved):

PR #227, opened 2026-08-14
fix: preserve completed results after cleanup failures

zhongxuanwang-nv:
#### Overview

Preserve a completed invocation result when runtime cleanup fails. The result is returned with a failed status and stop-stage error instead of being discarded.

#### Details

- Preserve successful `run_plan` output when `stop_runtime` fails.
- Apply the same behavior to Python's `Fabric.run` lifecycle.
- Add Rust and Python regressions.

#### Validation

- `cargo test -p nemo-fabric-core --lib` (102 passed)
- `cargo check -p fabric-python --locked`
- `uv run --no-sync pytest -q tests/python/test_runtime.py` (23 passed)
- `cargo fmt --all -- --check`; Ruff check; `git diff --check`
- The full `just test-rust` and `just test-python` commands could not complete within this terminal's foreground execution limit.

`crates/fabric-core/src/runtime.rs`: the cleanup-error branch in `run_plan` retains the completed result.

- Relates to #140 (replacement)

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved successful invocation output when runtime shutdown encounters an error.
  * Reported shutdown failures as structured, non-retryable errors instead of discarding completed results.
  * Prevented cleanup errors from replacing an invocation that had already failed.
  * Continued raising shutdown errors when no invocation result is available.

### DIFF
--- a/crates/fabric-core/src/runtime.rs
+++ b/crates/fabric-core/src/runtime.rs
@@ -712,7 +712,20 @@ pub fn run_plan(plan: &RunPlan, request: RunRequest) -> Result<RunResult> {
             return Err(error);
         }
     };
-    result.events.extend(stop_runtime(plan, &runtime)?);
+    match stop_runtime(plan, &runtime) {
+        Ok(events) => result.events.extend(events),
+        Err(error) if result.status == RunStatus::Succeeded => {
+            result.status = RunStatus::Failed;
+            result.error = Some(ErrorInfo {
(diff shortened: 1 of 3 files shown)


AjayThorve (approved):
Verified at cfbed891. The cleanup path preserves completed output, reports a structured stop-stage failure, and retains the original invocation failure when both phases fail. Focused Rust and Python regressions pass; current CI is green.

PR #230, opened 2026-08-14
chore: Improve slow Python tests

dagardner-nv:
#### Overview

* Removed `test_generated_python_scaffold_installs_editable` this test created a new venv and verified that the repo was installable as an editable workspace. Each run took about ~10s.
* Replace `test_adapter_python_data_path_query_times_out` with a Rust test.
  + The issue here is that we don't expose the adapter Python timeout time, which is hard-coded to 5s.
  + Created a impl method that exposes the timeout, and tested it in Rust with a 50ms timeout
* Re-work the api-server fixture to be session scoped cutting down on startup time

* `crates/fabric-python/src/lib.rs`

## Summary by CodeRabbit

- **Bug Fixes**
  - Added configurable timeouts for Python data-path queries, with clearer timeout error messages.
  - Ensured stalled processes terminate within the expected timeframe.

- **Tests**
  - Improved mock API server reset behavior between tests.
  - Reused a shared test server while preserving test isolation.
  - Removed obsolete Python scaffold and timeout regression tests.

### DIFF
--- a/crates/fabric-python/Cargo.toml
+++ b/crates/fabric-python/Cargo.toml
@@ -12,7 +12,6 @@ repository.workspace = true
 [lib]
 name = "_native"
 crate-type = ["cdylib", "rlib"]
-test = false
 doctest = false
 
 [dependencies]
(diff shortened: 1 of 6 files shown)


AjayThorve (approved):
LGTM

PR #244, opened 2026-08-17
ci: repair TypeScript package publication

AjayThorve:
#### Overview

Repairs two TypeScript adapter-contract publication defects:

- npm registry metadata omitted the package README because the helper published a prepacked tarball. npm therefore rendered unrelated repository content for `0.2.0-rc.2` instead of the package guide.
- nightly alpha tags are pushed with `GITHUB_TOKEN`, so GitHub suppresses the follow-on tag `push` workflow and never starts the npm publisher. Human-pushed RC tags do not have that limitation.

The change preserves trusted publishing through the protected `npmjs` environment and does not add a long-lived npm credential. Existing npm versions are immutable, so the README correction takes effect with the next published version; it cannot rewrite `0.2.0-rc.2`.

No breaking changes.

#### Details

- Publish the package directory with lifecycle scripts disabled so npm records `README.md` in registry metadata while uploading the same bytes that were prepacked and checked.
- Verify the registry README exactly matches the packaged README in addition to the existing version, integrity, and dist-tag checks.
- Remove obsolete pre-publication tarball instructions from the consumer README.
- Make `publish_typescript.yml` reusable with explicit tag-ref inputs while retaining direct `v*` tag pushes for stable, beta, and RC releases.
- Invoke that reusable publisher directly from `nightly-alpha-tag.yml` with the created alpha tag and minimum `contents: read` plus `id-token: write` permissions.
- Add workflow regression tests covering both RC/alpha tag selection and the nightly caller contract.

#### Validation

- `just test-python` — 1,191 passed, 18 skipped
- `just test-typescript`
- Focused publication and workflow tests — 38 passed
- Pre-commit checks on all changed files, including Ruff, copyright, and actionlint
- `just --fmt --check`
- `git diff --check`
- Compared `npm pack --ignore-scripts` with `npm publish . --dry-run --ignore-scripts`; file inventories and SHA-512 integrity matched exactly
- Confirmed the live `0.2.0-rc.2` registry metadata has an empty README and that RC tag pushes start the TypeScript publisher while nightly alpha runs do not

Start with `scripts/ci/publish_typescript_package.py` for the README metadata invariant, then review the caller/callee contract between `.github/workflows/nightly-alpha-tag.yml` and `.github/workflows/publish_typescript.yml`.

Relates to: none.

## Summary by CodeRabbit

* **New Features**
  * Nightly alpha releases can now automatically publish TypeScript packages.
  * TypeScript publishing supports reusable workflows with specified release tags.
  * Package publication now includes README metadata for registry verification.

* **Bug Fixes**
  * Prevented alpha tags from triggering standard release publishing.
  * Improved detection and reporting of missing or mismatched package README metadata.

* **Documentation**
  * Updated installation instructions to use the public npm registry.

* **Tests**
  * Added coverage for publishing workflows, nightly releases, and README verification.

### DIFF
--- a/.github/workflows/publish_typescript.yml
+++ b/.github/workflows/publish_typescript.yml
@@ -7,23 +7,13 @@ on:
   workflow_call:
     inputs:
       ref:
-        description: 'The Git ref to check out; defaults to the triggering commit'
-        required: false
-        default: ''
-        type: string
-      ref_name:
-        description: 'The effective ref name; defaults to the triggering ref name'
-        required: false
-        default: ''
-        type: string
-      ref_type:
-        description: 'The effective ref type; defaults to the triggering ref type'
-        required: false
-        default: ''
+        description: 'The Git tag ref to publish'
+        required: true
         type: string
   push:
     tags:
       - 'v*'
+      - '!v*-alpha*'
 
 concurrency:
   group: publish-typescript
@@ -36,7 +26,7 @@ defaults:
 jobs:
   publish-typescript:
     name: Publish (npmjs.com)
-    if: ${{ (inputs.ref_type || github.ref_type) == 'tag' }}
+    if: ${{ github.event_name == 'workflow_call' || github.ref_type == 'tag' }}
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     permissions:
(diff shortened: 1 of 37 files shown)


.github/workflows/publish_typescript.yml
```diff
@@ -20,7 +26,7 @@ defaults:
 jobs:
   publish-typescript:
     name: Publish (npmjs.com)
-    if: ${{ github.ref_type == 'tag' }}
+    if: ${{ github.event_name == 'workflow_call' || github.ref_type == 'tag' }}
```
zhongxuanwang-nv, line 29: My agent verified this that I'm think is also true; 

> 
> github.event_name is never 'workflow_call' inside a called workflow — it holds the caller's event. On the nightly path this evaluates to github.event_name == 'schedule' and github.ref_type == 'branch', so false || false skips the job. A skipped job reports green, so the nightly succeeds while publishing nothing.
> 
> Confirmed in actions/runner#3146 (https://github.com/actions/runner/issues/3146): a workflow_dispatch caller yields event_name=workflow_dispatch inside the callee (comment with screenshot).
> 
> This repo already uses the correct pattern — ci_python.yml:45 reads (inputs.ref_type || github.ref_type) != 'tag' and nightly-alpha-tag.yml:86 passes ref_type: tag for exactly this reason. Suggest if: ${{ inputs.ref != '' || github.ref_type == 'tag' }}, or pass an explicit input.

AnuradhaKaruppiah (approved):

.github/workflows/publish_typescript.yml
zhongxuanwang-nv: Oops! Sorry that my review took too long 😂 but i can address it elsewhere!

PR #253, opened 2026-08-25
feat: add remote agent adapter

dagardner-nv:
#### Overview

- Add the packaged Remote Agent adapter for terminal `invoke` over HTTP(S), supporting OpenAI Responses, OpenAI Chat Completions, and Anthropic Messages-compatible endpoints.
- Keep Fabric `invoke_stream` and Relay telemetry out of this PR; those remain follow-up work.

#### Details

- Map the supported Fabric configuration surface to remote requests: endpoint and protocol selection, model and credentials, temperature, Anthropic `max_tokens`, replacement system instructions, and bounded connect/read timeouts.
- Preserve a serialized per-runtime user/assistant transcript across invocations and return normalized output, usage, and retryability-aware failures.
- Declare the supported capability surface explicitly: terminal invocation only, with no Fabric streaming, updates, cancellation, service mode, or telemetry.
- Add package extras, lockfile and CI integration, installation and configuration documentation, and protocol/configuration/error-path tests.
- Reuse a shared input-normalization helper for adapters that accept structured input.

#### Validation

- `uv run --with pre-commit pre-commit run --all-files --show-diff-on-failure`
- `just --set no_uv true test-python` (1272 passed, 84 skipped)
- `uv run --no-sync pytest tests/adapters/test_remote_agent.py tests/python/test_harness_settings_validation.py` (55 passed)
- `just docs` (0 errors; unauthenticated redirect check skipped)
- `just build-python`
- `just wheels`
- `uv run --no-project python scripts/licensing/license_diff.py --base-ref origin/main`

- `adapters/remote-agent/src/nemo_fabric_adapters/remote_agent/adapter.py`
- `adapters/remote-agent/remote-agent.fabric-adapter.json`
- `tests/adapters/test_remote_agent.py`

- Closes FABRIC-223


adapters/remote-agent/src/nemo_fabric_adapters/remote_agent/adapter.py
```diff
@@ -0,0 +1,411 @@
...
+                    delta = value["delta"]
+                    if delta["type"] == "text_delta":
+                        text += delta["text"]
+                elif event == "message_delta":
+                    output_tokens = value["usage"].get("output_tokens")
+                elif event == "message_stop":
+                    return text, _usage(input_tokens, output_tokens, None)
+        raise RuntimeError("remote agent response ended without completion")
+
+    def _prepare_relay(
```
AnuradhaKaruppiah: for relay we may want to take a slightly different approach:
- we can't really control the relay config as the remote agent start is done outside of fabric
- fabric could provide an ATOF stream sync (invoke_stream support) if relay support has been enabled. This can be via config or via a new api to discover the remote agents capabilities.

adapters/remote-agent/src/nemo_fabric_adapters/remote_agent/adapter.py
```diff
@@ -0,0 +1,411 @@
...
+            self._runtime_id = context.runtime_id
+        except Exception:
+            try:
+                await self.stop()
+            except Exception: # preserve the original failure
+                pass
+            
+            raise
+
+    async def invoke(
```
AnuradhaKaruppiah, line 177: Could we support both terminal and native-streaming invocation modes? invoke() can continue consuming the response and returning a terminal AgentRunResult, while protocol-specific methods such as invoke_chat_completions_stream, invoke_responses_stream, and invoke_anthropic_messages_stream expose the corresponding native events and still provide a separate terminal result.
In the current implementation, Responses and Anthropic streams are consumed internally and reduced primarily to text and usage. This means the event sequence and protocol-specific content—such as tool calls, reasoning events, content-block structure, stop reasons, and other metadata—is not available to callers. Since preserving native streams is an important goal of this adapter, it would be useful to represent these protocols directly rather than map them all to the existing Chat Completions stream profile.

AnuradhaKaruppiah: This can be deferred to a separate PR

adapters/remote-agent/src/nemo_fabric_adapters/remote_agent/adapter.py
AnuradhaKaruppiah: an immediate follow up PR will be created to add invoke_stream support

AnuradhaKaruppiah (approved):

PR #260, opened 2026-08-31
feat: propagate UUID request IDs to Relay

AnuradhaKaruppiah:
#### Overview

Use a UUID Fabric `request_id` as the propagated Relay root for Deep Agents telemetry. When the request ID is not a UUID, preserve the existing correlation behavior by recording it as `nemo_fabric_request_id` metadata.

This is intentionally a Fabric-only adapter change: it does not add a session field or change the RunRequest, runtime-context, or adapter schemas.

#### Details

- Validate `runtime_context.request_id` with Python's UUID parser.
- For UUID values, create an isolated Relay scope stack whose parent and root are the request UUID, then open the existing Agent scope beneath it.
- For non-UUID values, keep Relay's default scope root and attach the request ID to the Agent scope metadata.
- Restore the prior Relay scope stack after every UUID request.
- Add unit coverage for both branches and an integration test against the real Relay propagation API.

With `session_id_source = "propagation_root"` from NVIDIA/NeMo-Relay#959, the same UUID is emitted as the ATIF `session_id` while Relay retains a distinct Agent/trajectory UUID.

#### Validation

- `uv run --no-sync pytest tests/adapters/test_deepagents.py tests/integrations/test_relay_scope_leak.py` — 80 passed
- `just --set no_uv true test-python` — 1,247 passed, 17 skipped
- Ruff format and lint checks on all three changed files
- `git diff --check`
- Cross-repository ATIF smoke test against NVIDIA/NeMo-Relay#959 confirmed `request_id == session_id` and a distinct `trajectory_id`

Not run: Rust and TypeScript suites because this change does not touch Rust, TypeScript, schemas, or native bindings.

Start with `_relay_request_context` and its use around the Agent scope in `adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py`. The main design decision is to treat a UUID request ID as an explicit propagation signal while retaining metadata fallback for existing non-UUID IDs.

- Relates to NVIDIA/NeMo-Relay#958

## Summary by CodeRabbit

* **New Features**
  * Relay telemetry now correlates requests using valid UUID request IDs and propagated request context.
  * Non-UUID request IDs continue to use request metadata for tracking.
  * Streaming requests now include their corresponding request identifiers in telemetry records.

* **Bug Fixes**
  * Improved telemetry scope handling and request correlation.
  * Scopes are restored after each request, preventing telemetry data from leaking between requests.

### DIFF
--- a/adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
+++ b/adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
@@ -963,12 +963,14 @@ def _join_faults(*faults: str | None) -> str | None:
 
 
 def _relay_request_context(request_id: str) -> tuple[Any, dict[str, str]]:
-    """Use a UUID request id as Relay's propagated root, or preserve it as metadata."""
+    """Use a UUID request id as Relay's propagated root and preserve its metadata."""
+
+    metadata = {"nemo_fabric_request_id": request_id}
 
     try:
         request_uuid = str(uuid.UUID(request_id))
     except ValueError:
-        return nullcontext(), {"nemo_fabric_request_id": request_id}
+        return nullcontext(), metadata
 
     from nemo_relay import PropagationContext
     from nemo_relay import create_scope_stack_from_propagation
@@ -976,7 +978,7 @@ def _relay_request_context(request_id: str) -> tuple[Any, dict[str, str]]:
 
     propagation = PropagationContext(request_uuid, root_uuid=request_uuid)
     stack = create_scope_stack_from_propagation(propagation)
-    return use_scope_stack(stack), {}
+    return use_scope_stack(stack), metadata
 
 
 def _relay_dependency_error() -> RuntimeError:
--- a/adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
+++ b/adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
@@ -18,6 +18,7 @@ import math
 import os
 import uuid
 from collections.abc import Callable
+from contextlib import nullcontext
 from pathlib import Path
 from typing import Any
 from typing import NamedTuple
@@ -689,7 +690,7 @@ class DeepAgentsRuntime:
     async def _invoke_with_telemetry(
         self,
         user_message: str,
-        request_id: str | None,
+        request_id: str,
     ) -> TurnOutcome:
         """Run one turn inside the Relay plugin/scope, isolating telemetry faults.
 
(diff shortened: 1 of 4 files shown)


tests/adapters/test_deepagents.py
```diff
@@ -262,6 +262,14 @@ async def plugin_ctx(config: object) -> AsyncIterator[dict[str, object]]:
     class ScopeType:
         Agent = "agent"
 
+    class PropagationContext:
+        def __init__(self, parent_uuid: str, root_uuid: str | None = None) -> None:
+            self.parent_uuid = parent_uuid
+            self.root_uuid = root_uuid
+            calls.setdefault("propagation_contexts", []).append(
+                (parent_uuid, root_uuid)
+            )
```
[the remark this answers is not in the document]
AnuradhaKaruppiah: Addressed in f843148c. The fixture now exposes `nemo_relay.PropagationContext` as a `MagicMock` with a side effect returning `types.SimpleNamespace`, while preserving the existing propagation call recording and assertions. Validation: focused test (2 passed), full Python suite (1247 passed, 17 skipped), and one-file pre-commit checks.

adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
```diff
@@ -959,6 +962,23 @@ def _join_faults(*faults: str | None) -> str | None:
...
+    except ValueError:
+        return nullcontext(), {"nemo_fabric_request_id": request_id}
+
+    from nemo_relay import PropagationContext
+    from nemo_relay import create_scope_stack_from_propagation
+    from nemo_relay import use_scope_stack
+
+    propagation = PropagationContext(request_uuid, root_uuid=request_uuid)
+    stack = create_scope_stack_from_propagation(propagation)
+    return use_scope_stack(stack), {}
```
yczhang-nv: Returning `{}` here removes the only turn-root marker that the SDK’s ATOF listener recognizes for the Deep Agents path. `invoke_stream()` arms the listener with `request_id` and `turn_index`, and `_belongs_to_active_turn()` drops records until `_matches_turn_root()` accepts a root. That predicate does not inspect the propagated parent/root UUID; its alternatives are Hermes-specific markers or `{nemo_relay_scope_role: "turn", turn_index: ...}`, and this adapter with Relay 0.7.3 does not emit `turn_index`.

Consequently, a UUID `request_id` can complete successfully while producing an empty ATOF stream. Could we return `{"nemo_fabric_request_id": request_id}` on both branches?

Please also update the adapter’s UUID metadata assertion and add an `invoke_stream` regression test using a UUID request ID. A listener-only test that manually supplies the metadata would not catch this producer-side regression.

yczhang-nv (approved):
Approved with one comment

adapters/deepagents/src/nemo_fabric_adapters/deepagents/adapter.py
AnuradhaKaruppiah: > {"nemo_fabric_request_id": request_id} on both branches?

Agree with this, that is the right approach even in a healthy ATOF case.

> Please also update the adapter’s UUID metadata assertion and add an invoke_stream regression test using a UUID request ID. A listener-only test that manually supplies the metadata would not catch this producer-side regression.
Will do

AnuradhaKaruppiah: Addressed in b6a772e4. UUID request IDs still seed Relay propagation, and `nemo_fabric_request_id` is now retained on the Agent scope for ATOF turn-root matching. I updated the adapter and real-Relay metadata assertions and changed the Deep Agents `invoke_stream` e2e path to use UUID request IDs and verify the emitted metadata. Validation: focused adapter tests (2 passed), UUID invoke_stream e2e (1 passed), real-Relay integration (1 passed), full Python suite (1247 passed, 17 skipped), and pre-commit checks.

dagardner-nv (approved):

Issue #297, opened 2026-09-15
[Bug]: Pi adapter cannot retain and resume conversation state across runtime restarts


cv:
## Affected area
Harness adapter; Consumer integration

## NeMo Fabric version
Source commit `51a28c1aefec56abd877070b6973d0a32a1e3003`.

## Installation method
Fabric Python SDK and TypeScript Pi adapter built from source in a local test container, image ID `sha256:69a045d358a5ef73cb4e448aa9555c2d198272445a39e80981145c4acffc4e1b`.

## Current behavior
The Pi adapter retains conversation context between invocations of one runtime, but loses that context when the runtime stops. Starting another runtime with the same configuration, workspace, and artifact directory does not restore it.

`adapters/typescript/pi/src/pi-sdk.ts:504` unconditionally passes `sessionManager: pi.SessionManager.inMemory(workspace)` to `createAgentSession`. The adapter README documents this in-memory behavior; this report concerns its limitation for persistent-agent consumers.

## Expected behavior
Provide an opt-in disk-backed session and an explicit session identity/resume mechanism, so consumers can continue a conversation after runtime restart. Sharing a workspace alone need not imply that separate runtimes share a conversation.

## Steps to reproduce
Save the complete reproduction below as `repro.py`. It runs the real Fabric SDK and packaged Pi adapter against a local OpenAI Responses fixture. The fixture records incoming model requests and returns a constant successful reply; no model recall judgment is involved.

Prerequisites: install the Fabric Python SDK and build the TypeScript Pi adapter with its Pi peer dependencies at the source revision above, following this repository's setup instructions. Run with a supported Node.js version on PATH and the Fabric SDK in the selected Python environment.

```sh
OPENAI_API_KEY=fixture-not-a-secret \
ADAPTER_PYTHON="$(command -v python)" \
python repro.py /absolute/path/to/NeMo-Fabric/adapters/typescript/pi/pi.fabric-adapter.json
```

The script uses a temporary working directory and a loopback HTTP fixture. It needs no external inference service. The observed run used a local source-built container with networking disabled; the command above makes the source descriptor and Python environment explicit instead of requiring that container image.

The essential SDK sequence is:

```python
# config selects nvidia.fabric.pi; root and workspace/artifact paths stay fixed across both starts.
runtime = await Fabric().start_runtime(config, base_dir=str(root))
try:
    await runtime.invoke(input='Remember SESSION_MARKER_7a943c')
[...]

PR #303, opened 2026-09-15
fix(pi): resume persistent sessions

yczhang-nv:
#### Overview

Add explicit disk-backed session creation and resume to the Pi adapter so a caller can continue a conversation after stopping one NeMo Fabric runtime and starting another.

The existing behavior remains the default: omitting `harness.settings.session` creates an isolated in-memory session. Persistent sessions are opt-in, caller-named, and stored under a workspace-relative directory. This change adds no dependencies and has no breaking changes.

#### Details

- Add validated `create` and `resume` modes under `harness.settings.session`.
- Use exclusive file creation so duplicate session IDs fail without overwriting existing history, and release the reservation on every failed startup path.
- Pass the runtime workspace explicitly to Pi's persistent session manager so session headers do not depend on the adapter process working directory.
- Reject unsafe session directories during planning and retain runtime containment and symlink checks.
- Reject missing, empty, malformed, or inaccessible session files with stable Pi adapter error codes.
- Document that one persistent session ID has a single live-runtime writer and concurrent resume is unsupported.
- Document the trusted-storage requirement while a Pi runtime is active because the upstream SDK persists through pathname-based file operations.
- Build the packaged Pi adapter in every Python CI matrix job and fail collection if CI would otherwise skip the resume regression.
- Add descriptor, workflow, process-host, malformed-session, and end-to-end coverage for persistence and default runtime isolation.

#### Validation

- `PYO3_PYTHON=.venv/bin/python just test-rust` — passed, including 121 core tests; generated schema and contract outputs remained clean.
- `just test-python` — 1,404 passed, 18 skipped; the Pi E2E ran and passed.
- `just test-typescript` — passed, including Pi's 21 unit/process tests plus contract generation, types, package, install, dependency, and audit checks.
- `.venv/bin/python -m pytest -q tests/e2e/test_pi.py` — 1 passed.
- `just docs` — passed; the authenticated Fern missing-redirect check was skipped because credentials were unavailable.
- Changed-file pre-commit hooks, `just --fmt --check`, and `git diff --check` — passed.

Start with the Python job in `.github/workflows/ci_python.yml` and the CI assertion in `tests/e2e/test_pi.py`, which make the successful resume regression mandatory. Then review `createSessionManager` and `validateSessionFile` in `adapters/typescript/pi/src/pi-sdk.ts` for explicit workspace cwd, strict JSONL validation, and reservation cleanup semantics.

- Relates to: #297

## New Features
- Added explicitly configured persistent Pi sessions with create and resume modes.
- Supports workspace-relative storage directories and validated session IDs.
- Sessions remain in memory by default; persistence requires explicit configuration.
- Added validation and handling for duplicate, missing, malformed, inaccessible, unsupported, and symlinked sessions.
- Added restrictions for concurrent access and unsafe storage paths.

## Documentation
- Documented session configuration, lifecycle behavior, storage restrictions, validation rules, and failure codes.

## Tests
- Added coverage for persistence, isolation, configuration validation, and lifecycle failures.


adapters/typescript/pi/src/pi-sdk.ts
```diff
@@ -140,6 +168,107 @@ function isRecord(value: unknown): value is Record<string, unknown> {
...
+      await placeholder.close();
+    } catch (error) {
+      if (errorCode(error) === "EEXIST") {
+        throw new LifecycleError("pi_session_exists", "A Pi session with the configured id already exists");
+      }
+      throw new LifecycleError("pi_session_storage_failed", "The Pi session file could not be created");
+    }
+  } else {
+    try {
+      const info = await stat(sessionFile);
```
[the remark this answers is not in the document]
yczhang-nv: Fixed in b5c0162. Resume now uses lstat and rejects a final-component session-file symlink before passing the path to SessionManager.open. The process-host regression creates a valid session outside the workspace, links it into the session directory, and verifies pi_session_directory_outside_workspace.

adapters/typescript/pi/src/pi-sdk.ts
```diff
@@ -501,8 +631,8 @@ export class PiSdkSessionFactory implements PiSessionFactory {
       model,
       modelRuntime,
       resourceLoader,
-      sessionManager: pi.SessionManager.inMemory(workspace),
-      settingsManager: settings,
+      sessionManager: await createSessionManager(pi, workspace, settings.session),
```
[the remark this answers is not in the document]
yczhang-nv: Fixed in b5c0162. Session preparation now returns a cleanup callback tied to the reserved file identity. The startup error boundary stops any created handle and removes only that reservation; the regression retries the same session ID after a pi_tool_missing startup failure and receives the original error again instead of pi_session_exists.

adapters/typescript/pi/src/pi-sdk.ts
```diff
@@ -140,6 +168,153 @@ function isRecord(value: unknown): value is Record<string, unknown> {
...
+      if (errorCode(error) === "ENOENT") {
+        throw new LifecycleError("pi_session_not_found", "The configured Pi session does not exist");
+      }
+      throw new LifecycleError("pi_session_storage_failed", "The Pi session file could not be read");
+    }
+  }
+
+  try {
+    return {
+      sessionManager: pi.SessionManager.open(sessionFile, sessionDirectory, workspace),
```
sara-tadayon-nv, line 522: The Fabric contract documents that `resume` on a malformed session should return `pi_session_invalid` but we are only validating file metadata before passing the file to `pi.SessionManager.open()`. Pi’s parser can skip malformed JSONL records, so a partially corrupt session could resume successfully with missing context instead of returning the documented error.

Before opening the session, we should add file-content validation for this behavior:
- Read the file as UTF-8 and allow trailing blank lines only.
- Require every non-blank line to parse as JSON; do not silently skip malformed lines.
- Require the first record to be a valid Pi session header, using a Pi-supported session version.
- Validate that subsequent records have the required shape for their Pi entry type, especially message entries.
- Return LifecycleError("pi_session_invalid", ...) for any invalid record.

Tests to add:
- A valid header followed by truncated JSON.
- A valid header followed by an incomplete entry such as {"type":"message"}.
- A valid session with trailing blank lines, which should still resume.

sara-tadayon-nv, line 522: We validate that `sessionFile` is not a symlink before this call, but Pi continues to read and write through this pathname after startup. The file could be replaced after validation, and a later Pi write could follow a symlink outside the workspace.
Is the workspace intended to be trusted for the lifetime of a runtime? If not, we might need a persistence mechanism that keeps the validated file identity/no-follow behavior for subsequent writes.

yczhang-nv: Thanks! Fixed in `8b733361`, which performs the validation you mentioned before opening the session. Test coverages are added as well.

yczhang-nv: Yeah, the adapter can not guarantee Pi won't follow a symlink that's replaced later which is outside the workspace. The limitation is on Pi's implementation, briefly:

- `SessionManager._rewriteFile()` reopens the pathname with "w".
- `SessionManager._persist()` repeatedly calls `appendFileSync(path, ...)`.
- `createAgentSession`  accepts the concrete `SessionManager`, whose constructor is private, rather than a persistence interface.

I think the most preferred solution here is to add an upstream Pi `SessionStore` interface to add an FD-backed implementation. For the scope of this PR, I added documentation in `8b733361` to mention that the session directory is trusted runtime storage, and callers must not replace the directory or session file while a runtime is using it. Let me know if that looks good to you!

sara-tadayon-nv (approved):
lgtm!

adapters/typescript/pi/README.md
```diff
@@ -85,6 +87,50 @@ harness = HarnessConfig(adapter_id="nvidia.fabric.pi")
...
+After the runtime starts successfully and stops normally, start a new runtime
+with the same configuration except for `"mode": "resume"`. A session is never
+selected only because two runtimes use the same workspace.
+
+The default session file is
+`<workspace>/.fabric-pi/sessions/<id>.jsonl`. An optional `directory` selects a
+different workspace-relative directory. The adapter rejects absolute paths,
+parent-directory escapes, session-file symlinks, and directory symlinks that
+resolve outside the workspace.
+Session IDs may contain letters, numbers, `.`, `_`, and `-`; they must start and
```
lvojtku, line 116:
```suggestion
Session IDs may contain letters, numbers, `.`, `_`, and `-` and must start and
```

adapters/typescript/pi/README.md
```diff
@@ -85,6 +87,50 @@ harness = HarnessConfig(adapter_id="nvidia.fabric.pi")
...
+JSONL path when it persists session updates and does not provide a no-follow
+file-descriptor API.
+
+`create` returns `pi_session_exists` rather than overwriting an existing file.
+`resume` returns `pi_session_not_found` for a missing file and
+`pi_session_invalid` for an empty or malformed file. Other storage or permission
+failures return `pi_session_storage_failed`. Stopping a runtime preserves the
+file. Use only one live runtime per persistent session ID. Concurrently resuming
+the same session is unsupported because Pi does not lock its JSONL session file.
+The caller owns retention: after the runtime stops, delete
```
lvojtku, line 130:
```suggestion
The caller owns retention, where after the runtime stops, it delete
```

docs/integrations/harness/pi.mdx
```diff
@@ -85,6 +85,59 @@ the environment variable that contains the provider credential.
...
+
+Treat the session directory as trusted runtime storage. Do not replace the
+directory or session file while a runtime is using it. The Pi SDK reopens the
+JSONL path when it persists session updates and does not provide a no-follow
+file-descriptor API.
+
+Stopping a runtime preserves its session file. Use only one live runtime per
+persistent session ID. Concurrently resuming the same session is unsupported
+because Pi does not lock its JSONL session file. The caller owns retention:
+delete `<directory>/<id>.jsonl` after the runtime stops when the conversation is
```
lvojtku, line 137:
```suggestion
because Pi does not lock its JSONL session file. The caller owns retention, where
delete `<directory>/<id>.jsonl` after the runtime stops when the conversation is
```

lvojtku (approved):
Approved with comments

PR #306, opened 2026-09-16
feat: name the model credential variable for Harbor FabricAgent runs

SandyChapman:
#### Overview

`FabricAgent` generates the `FabricConfig` for a Harbor run from Harbor's inputs, but the model config it builds has no `api_key_env`. Adapters that require an explicit credential variable for non-OpenAI providers (the deepagents adapter does) therefore cannot run against an endpoint such as build.nvidia.com from Harbor: there is no kwarg that reaches the credential name, and `fabric_harness_settings` cannot override model config.

This adds `fabric_model_api_key_env` on `FabricAgent`, threaded through `build_harbor_config` onto `models.default.api_key_env`. It is rejected without a model, matching `fabric_model_base_url`, and rejected when blank. The Harbor examples README documents the kwarg and how to pair it with `--ae <NAME>=<value>` so the variable exists in the task container.

```bash
harbor run --agent nemo_fabric.integrations.harbor:FabricAgent \
  --model nvidia/nemotron-3-super-120b-a12b \
  --ak fabric_adapter_id=nvidia.fabric.langchain.deepagents \
  --ak fabric_model_base_url=https://integrate.api.nvidia.com/v1 \
  --ak fabric_model_api_key_env=NVIDIA_API_KEY \
  --ae "NVIDIA_API_KEY=$NVIDIA_API_KEY" ...
```

Motivation: NeMo Platform's evaluator runs Fabric harnesses inside its Harbor runner and delivers credentials as environment variables; without this it carries a `FabricAgent` subclass just to set `api_key_env`.

Related: #307 fixes the deepagents adapter sending Fabric's `openai/` slug prefix to OpenAI. The two are independent, but a deepagents run on a Nemotron model through Harbor needs both.

**Testing**
- `uv run pytest tests/python/test_harbor_integration.py` → 22 passed, including three new tests: the credential name and endpoint land on the generated model config, a credential name without a model is rejected, a blank name is rejected at construction.
- `ruff check` clean; the new hunk is `ruff format`-clean.

No breaking changes: the kwarg defaults to `None` and existing configs are unchanged.

`build_harbor_config` in `sdk/python/nemo-fabric-runtime/src/nemo_fabric/integrations/harbor/fabric_agent.py`, the `ModelConfig` construction.

- Relates to #307
- Relates to NVIDIA-NeMo/nemo-platform#2079

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the model’s API key environment variable in Harbor integrations.
  * Model endpoint and credential settings are now included in generated Harbor configurations.
  * Added guidance for passing the corresponding container environment variable.

* **Bug Fixes**
  * Invalid configurations are rejected when the credential name is blank or provided without a model name.

* **Tests**
  * Added coverage for model credential and endpoint configuration scenarios.


sdk/python/nemo-fabric-runtime/src/nemo_fabric/integrations/harbor/fabric_agent.py
```diff
@@ -104,6 +104,7 @@ def __init__(
             fabric_workspace: str = HARBOR_DEFAULT_WORKSPACE,
             fabric_harness_settings: dict[str, Any] | None = None,
             fabric_model_base_url: str | None = None,
+            fabric_model_api_key_env: str | None = None,
```
AnuradhaKaruppiah: Adding this parameter here changes the binding of every subsequent positional argument. Because `FabricAgent.__init__` has no keyword-only separator, an existing caller that passes `fabric_system_instruction` positionally will now silently bind that value to `fabric_model_api_key_env`, with later arguments shifted as well. This contradicts the PR’s backward-compatibility claim.

Please preserve the existing positional signature by making the new option keyword-only—for example, place `fabric_model_api_key_env` after `*args` and before `**kwargs`. `build_harbor_config` is already keyword-only, so it does not have the same issue.

SandyChapman: moved `fabric_model_api_key_env` after `*args` so it's keyword-only, and added a test

AnuradhaKaruppiah (approved):

PR #313, opened 2026-09-21
chore: upgrade to Relay 0.9 and Hermes with Relay 0.9 support

marcusds:
> **Draft.** Blocked on NousResearch/hermes-agent#115343 (Hermes upgrade to Relay 0.9.0). Hermes is pinned at that PR's head commit until it merges and Hermes cuts a release; the pin then moves to the release tag. Everything else is ready for review.

#### Overview

Upgrade NeMo Relay from `>=0.7.2,<0.8` to `>=0.9,<0.10`, and move the Hermes Agent pin to a commit with Relay 0.9 support. This is the minimum set of changes needed to run on Relay 0.9. Relay 0.9 features (session-rooted propagation, typed `header_file`) and cleanup of code Relay 0.8 made dead are deliberately left to follow-up PRs stacked on this one.

Relay 0.9 forces four changes, per the [migration guide](https://docs.nvidia.com/nemo/relay/reference/migration-guides#upgrade-to-nemo-relay-09) and [release notes](https://github.com/NVIDIA/NeMo-Relay/releases/tag/0.9.0):

- `plugin.plugin()` is gone. Adapters use `plugin.activate()` and read `activation.report`, whose diagnostics now nest under `config`.
- Tool execution intercepts must return `ToolExecutionResult`.
- Subscriber callbacks are enqueued, so Hermes' ATIF file lands after the session scope pops. The Hermes adapter flushes subscribers after `finalize_session`.
- Overlapping LangGraph runs no longer strand a scope. The scope-leak integration tests assert the restored stack instead.

Hermes side: the pinned commit includes the `metadata-propagate.patch` change upstream (NousResearch/hermes-agent@6ee218602d), so the vendored patch and its justfile apply/verify logic are deleted. Current Hermes also titles every session with an auxiliary model call, which doubled ATIF steps; the generated Hermes config disables it.

Dependency changes and license diff (`scripts/licensing/license_diff.py --base-ref main python`):

- `nemo-relay` 0.7.2 → 0.9.0 (Apache-2.0), `nemo-relay-cli-bin` 0.7.2 → 0.9.0 (Apache-2.0).
- `hermes-agent` pinned at `47edd6455bbb66ac94e00be0ca3e888680f02882`, head of NousResearch/hermes-agent#115343. Declared version is still 0.21.3 because Hermes main has not cut a release since v2026.9.14; the released 0.21.3 does not contain Relay 0.9 support.
- New transitive: `pillow-heif` 1.7.0 (BSD-3-Clause), pulled in by current Hermes. Its wheels bundle `libheif` and codec libraries distributed under LGPL terms, which the lockfile inventory does not surface. Flagging for dependency approvers.
- `ATTRIBUTIONS-Python.md` regenerated.

Breaking for users: Claude, Codex, and Pi now share one CLI range, `>=0.9,<0.10`. The install guide's separate Claude/Codex and Pi install blocks collapse into one and the disjoint-range warning is removed.

`adapters/python/common/src/nemo_fabric_adapters/common/utils.py`, `reject_inherited_relay_plugin_config`. This is the one place the 0.9 report shape matters, and every adapter routes through it. Then `adapters/python/hermes/src/nemo_fabric_adapters/hermes/telemetry.py` for the subscriber flush, which is the only new runtime behaviour rather than a rename.

For the Hermes side, `justfile` `install-hermes-agent` shows what the upstream patch landing makes deletable.

- Relates to: NousResearch/hermes-agent#115343 (blocks merge; pin moves to the Hermes release containing it)


adapters/python/common/src/nemo_fabric_adapters/common/relay_gateway.py
```diff
@@ -21,8 +21,8 @@
 RELAY_HEALTH_TIMEOUT_SECONDS = 10.0
 RELAY_STOP_TIMEOUT_SECONDS = 5.0
 RELAY_VERSION_TIMEOUT_SECONDS = 5.0
-RELAY_MINIMUM_VERSION = (0, 7, 2)
-RELAY_MAXIMUM_VERSION = (0, 8, 0)
+RELAY_MINIMUM_VERSION = (0, 9, 0)
```
AnuradhaKaruppiah, line 24: What is the reason for bumping the min? We should allow the wider range unless there is a compelling reason to not do that

marcusds: Relay 0.9 introduces some backwards incompatible changes.

